Issue #210, opened 2015-01-16
Enable channels to be imported and exported as a json file


rcrichton:

rcrichton:
See the linked PR above for more description about what this issue entails.

PR #221, opened 2015-02-09
OHM-24: first version of export/import feature

BMartinos:
This is the first working version of the export/import features. This PR relates to issue: https://github.com/jembi/openhim-console/issues/210

### DIFF
--- a/app/scripts/services/rest.js
+++ b/app/scripts/services/rest.js
@@ -27,7 +27,6 @@ angular.module('openhimWebui2App')
       Transactions: $resource( server + '/transactions/:transactionId', { transactionId: '@_id' }),
 
       Mediators: $resource( server + '/mediators/:urn', { urn: '@urn' }),
-      MediatorsImport: $resource( server + '/mediators'),
 
       // add the metric endpoints
       Metrics: $resource( server + '/metrics/:type/:channelId', {}),
--- a/app/scripts/services/rest.js
+++ b/app/scripts/services/rest.js
@@ -9,11 +9,6 @@ angular.module('openhimWebui2App')
     var port = angular.copy( $rootScope.port );
     var server = protocol + '://' + host + ':' + port;
 
-    // delete API server details from rootScope
-    delete $rootScope.protocol;
-    delete $rootScope.host;
-    delete $rootScope.port;
-
     return {
       Authenticate: $resource( server + '/authenticate/:email' ),
 
@@ -32,6 +27,7 @@ angular.module('openhimWebui2App')
       Transactions: $resource( server + '/transactions/:transactionId', { transactionId: '@_id' }),
 
       Mediators: $resource( server + '/mediators/:urn', { urn: '@urn' }),
+      MediatorsImport: $resource( server + '/mediators'),
 
       // add the metric endpoints
       Metrics: $resource( server + '/metrics/:type/:channelId', {}),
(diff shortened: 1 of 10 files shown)


BMartinos:
Hey @rcrichton We you get some time can you review this PR for me. This is still a first version so a few small things would prob change. Ive also notice a few core issue when doing imports. Duplicates and handling issues. Will document them as well

app/scripts/services/rest.js
```diff
@@ -57,4 +52,35 @@ angular.module('openhimWebui2App')
...
+        update: { method: 'PUT' }
+      }),
+
+      Mediators: $resource( server + '/mediators:mediatorId/', { mediatorId: '@_id' }, {
+        update: { method: 'PUT' }
+      }),
+
+      ContactGroups: $resource( server + '/groups/:groupId', { groupId: '@_id' }, {
+        update: { method: 'PUT' }
+      }),
```
rcrichton: Why do you need a new factory for these? Can you use the existing resources above? Do these differ is some way?

app/views/exportImport.html
```diff
@@ -0,0 +1,130 @@
...
+          </div>
+        </div>
+
+        <!-- Top Alerts -->
+        <alert ng-repeat="alert in alerts.top" type="alert.type">{{alert.msg}}</alert>
+        <!-- Top Alerts -->
+        
+        <div class="exportSectionDiv">
+
+          <h3><strong>Export your configurations</strong></h3>
```
rcrichton: Maybe remove the word 'your' so it just reads 'Export Configurations'

app/views/exportImport.html
```diff
@@ -0,0 +1,130 @@
...
+          </div>
+        </div>
+
+
+
+        
+
+
+        <div class="importSectionDiv">
+          <h3><strong>Import your configurations</strong></h3>
```
rcrichton: As above maybe change to 'Import Configurations'

app/views/exportImport.html
```diff
@@ -0,0 +1,130 @@
...
+            <div class="btn-group" style="margin-bottom: 10px;">
+              <label class="btn btn-primary" ng-model="exportSettings.removeIds" btn-radio="true"><strong>Insert</strong> Script</label>
+              <label class="btn btn-primary" ng-model="exportSettings.removeIds" btn-radio="false"><strong>Update</strong> Script</label>
+            </div>
+          </div>
+          
+          
+          <div class="exportCollectionsDiv" ng-repeat="(model, collectionData) in exportCollections">
+          
+            <label class="exportModelTitle">All {{model}} <input type="checkbox" ng-checked="true" ng-model="checkBox" value="{{collectionData}}" ng-change="toggleCollectionExportSelection(model, collectionData)" /></label>
```
rcrichton: I think moving the checkboxes to the front of the text will look better. For both the All {{model}} and the individual records.

rcrichton:
@BMartinos this is looking good. A few comments above. We also need some unit tests for the major functions that you have created.

app/scripts/services/rest.js
BMartinos: I had some issues with the API not being consistent with the imports. Namely Users and Mediators, They do not use '_id' as their reference key so the imports would always fail because for Users e.g email would always be attached to the endpoint which is not what needs to happen for the import. 

I created the separate import factory as the current endpoints were not all compatable with the import process and to separate the diff kind ofendpoints and it allows for more control over the import

app/views/exportImport.html
BMartinos: cool will update

app/views/exportImport.html
BMartinos: cool will update

app/scripts/services/rest.js
rcrichton: Ok I see. I'd prefer to keep this to one API but if that's not possible then I think you have a good solution here.

BMartinos:
hey @rcrichton i have completed the changes and added some unit test. Can you review this for me again please. Thanks

app/views/exportImport.html
```diff
@@ -0,0 +1,128 @@
...
+
+        <!-- Top Alerts -->
+        <alert ng-repeat="alert in alerts.top" type="alert.type">{{alert.msg}}</alert>
+        <!-- Top Alerts -->
+        
+        <div class="exportSectionDiv">
+
+          <h3><strong>Export Data</strong></h3>
+          <div style=" float: left; clear: both; width: 100%; margin-bottom: 10px;">
+            <label tooltip="For a 'Insert Script' all '_id' properties will be removed and re-created on import">Will this export file be used as an Insert or Update?</label><br />
```
rcrichton: Maybe for a tooltip we should speak to something the user will understand. Something like: "For an 'Insert Script' the exported data will always add new elements, however, an 'Update Script' will attempt to update existing elements if possible"

rcrichton:
Hey @BMartinos look great :+1: Just one minor comment.

BMartinos:
Hey @rcrichton Ive made that small change as mentioned. I just needed to change something in core to avoid ContactGroup duplications. Ive created a PR here https://github.com/jembi/openhim-core-js/pull/357 if you can maybe just have a quick look at that as well.

rcrichton:
Awesome looks good :+1:

PR #233, opened 2015-02-22
Debian packaging

litlfred:
Seems to be working well on a clean Ubuntu 14.04 machine


packaging/targets/trusty/etc/init/openhim-console.conf
```diff
@@ -0,0 +1,22 @@
...
+
+
+setuid openhim
+setgid openhim
+
+env PATH=$PATH:/home/openhim/.nvm/current/bin
+
+script
+  chdir /home/openhim/openhim-console
+  exec /home/openhim/openhim-console/node_modules/grunt-cli/bin/grunt serve
```
rcrichton: We need to change this to not use grunt but rather copy the console files into Apache or another more production ready http-server. Copying into Apache seems easiest but how do we ensure that it doesn't overwrite anything? Should be expose it in an /openhim folder?

packaging/create-deb.sh
```diff
@@ -0,0 +1,181 @@
...
+    done
+    cd $HOME
+    
+    $GIT  status
+
+    $GIT add .
+
+    echo "Incrementing version"
+    $GIT commit ./ -m "\"${COMMITMSG}\""
+    $GIT tag $TAG
```
rcrichton: We are already tagging releases for our npm releases, could we not make use of those rather?

rcrichton:
Hey @litlfred I just have a few questions about this PR. I'd like to try simplify the deb if possible. Please let me know your thoughts.

rcrichton:
@devcritter could you review this for me please? I'm also wondering if I should squash the release commits in this branch. What do you think?

litlfred:
Just and FYI.. the packaging script looks at git tags to determine the changes since the last release.  It is used to update the changelog:
   https://github.com/jembi/openhim-console/blob/debian-packaging/packaging/targets/trusty/debian/changelog
If you squash the commits, that may make it a bit wonky, but you should be able to fix it by manually adding the tag "ubuntu-1.0.0-25" to get things started.

However, I was talking with Paul B and they said they wanted to have both stable and unstable packages.   So maybe we should include the PPA name in the tags like so:
  ubuntu-release-1.1.0-1
and  
  ubuntu-testing-1.1.0-1

The former would only include commits up to the "v1.0.0" tagged commit while the later would include everything.   

We of course we need to update the bash script to accommodate something like this.  I would be happy to walk through it.

rcrichton:
Thanks Carl, I've changed the scripts a fair bit so that it doesn't create its own tags so we can use our existing release tags that we create when making a github release and it doesn't add commits logs to the changelog. We already maintain a more readable changelog on github releases so I'd rather point to that. So I think squashing should be fine.

The unstable releases are interesting, this would require more discussion I think. Like how often should they be built, etc. We try to release stable releases as quickly as possible so maybe unstable releases aren't needed.

packaging/targets/trusty/debian/install
```diff
@@ -0,0 +1 @@
+/usr/share/nginx/html/*
```
devcritter: +1 for nginx :) Maybe it'd be better to put the console somewhere else though, in case the implementer is already using nginx and have existing content in there?

What we could do is add a entry for the console in `/etc/nginx/sites-available/` pointing to another location (maybe `/usr/share/openhim-console`?), then enable this in site-enabled. What do you think?

rcrichton: That's a good suggestion. Right now it get installed to /usr/share/nginx/html/openhim/ so that is doesn't conflict. But do you think that installing to its own place is better?

devcritter: Yeah I think so

rcrichton: Ok, cool. I'l make those changes.

litlfred:
Talking with Paul B., they want to have an unstable release on every commit.
Cheers,
-carl

rcrichton:
@litlfred ok I see. We will have to think about adding this in the future.

@devcritter could you review this again for me? I've changed the nginx config to use sites-enables.

devcritter:
:+1: Looks good to me!

PR #234, opened 2015-02-25
Switched form ngmin to ngAnnotate

### LINKED ISSUE #232: Switch from ng-min to ng-annotate
ng-min is deprecated and causing issues with uglify mangling. It looks like switching to ng-annotate will fix the problem.

I have temporarily disabled mangling in #231. Once this issue has been closed we should re-enable it.

rcrichton:
Closes #232 

@bausmeier can you have a quick look at this?


bausmeier:
:+1: Fixes the issue I was having previously.

PR #238, opened 2015-02-25
Updated Site Style - added Logo

BMartinos:
Added OpenHIM logo and updated header/footer styling


BMartinos:
Hey @rcrichton , @devcritter , @debonair I had a few spare minutes while waiting for feedback so quickly looked at adding the openHIM logo. Needed to change the header/footer slightly for the logo to work ( got lost in the dark background )

BMartinos:
With having our logo already on the opemhim.org i though it would be good to incorporate into the console. I didn't spend to much time on so buttons etc will still be blue. this is just a concept. changing button styles etc should be quick as these are only style colors

devcritter:
I think it looks great! I'm in favour of switching before the 1.1.0 release if possible.

rcrichton:
Very nice! I agree we definitely would like this for the 1.1 release. I actually think the blue button look at as they are. It may be worth experimenting with the green though.

rcrichton:
Is there any chance that you could make the footer always display at the bottom of the screen unless the page needs to scroll then rather just have it at the page so you only see it when you scroll to the end. Does this make sense? I just want to get away from the footer appearing in the middle of the screen.

BMartinos:
Ive added some green color. Ive gone with the darker green as the main one but is easy to change around. Links ive kept as blue as i think thats a general color for links. Some other places i haven't changed the color yet. Will get to those

BMartinos:
@rcrichton this might be a lil difficult but ive got a few ideas that i want to try that might work. The problem with this is its hard to tell how long the page is and whether it needs to be fixed or relative to the page. I can write a calculation to get the page height but this then needs to be adjusted according to screen size etc.

rcrichton:
Ok cool, the footer edits aren't to serious for now.

BMartinos:
Hey @rcrichton ive added the fixed footer in quick. So if the page can scroll then the footer is right at the bottom (possibly hidden based on content size) and If there isnt much content then the footer is fixed at the bottom of the screen

(P.S the code is a bit horrible as i just hacked it together quick to show the concept)

devcritter:
Another quick one: we should maybe update the copyright at the bottom of the page:
```
Copyright 2014-2015 Jembi Health Systems NPC
```
should be correct right?

BMartinos:
Yes we need to update the copyright to reflect the correct date. We can either have both start and current year or just the current year. Ill add a JS date to reflect this automatically

rcrichton:
Cool, I think just the current year makes sense. Also maybe you could make the default config to say 'Admin Console' instead of 'OpenHIM Admin Console' at the top. Else we are it seems like duplication with the logo.

rcrichton:
Otherwise, I think this is good to go for the release! :+1: What do you think @devcritter

devcritter:
Definitely! :+1:

BMartinos:
Great will just clean up the code and put the scripts in the correct places and upload again

rcrichton:
@BMartinos sorry, one more tiny thing. The little down arrow on the by the logged in users name. Can we change that to black?

BMartinos:
Hey @devcritter Ive made some changes just to clean-up my code. Ive also added the color changes to a separate css file which can be interchangeable ( future feature possibly ).

BMartinos:
@rcrichton and @debonair if you guys just maybe want to have a look and mention anything that needs to change

rcrichton:
Look god to me, thanks @BMartinos :+1:

debonair:
@BMartinos looks good to me too!

devcritter:
Same here! :+1:

PR #239, opened 2015-02-26
Fixed ngInclude reference for the user settings options

BMartinos:
Found an issue with the Profile page user settings. I moved the user-settings include script into the partials folder but didnt update the link for the profile page.

### DIFF
--- a/app/views/profile.html
+++ b/app/views/profile.html
@@ -94,7 +94,7 @@
 
               
               <!-- include for user settings tabs -->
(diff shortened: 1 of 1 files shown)


BMartinos:
Hey @devcritter just a quick fix for the profile page. I moved the user settings include to the partials folder and updated the users section but forgot about the profile page

devcritter:
:+1:

PR #244, opened 2015-02-26
Changed the favicon to our new logo, removed unused images.

rcrichton:
@BMartinos and @devcritter just changed the favicon and got rid of so unused images. This look ok?


devcritter:
:+1: looks good to me!

BMartinos:
Ive got a smaller icon version if you want to use that but otherwise looks good to me :+1:

rcrichton:
I think its ok at that size since it will be cached.

PR #245, opened 2015-02-26
Channel load fix

BMartinos:
Fixed an issue with channel load results including an extra day. Had to do date validation to only add if timestamp falls within past 7 days (current day included)


BMartinos:
Hey @rcrichton do you mind just reviewing this fix for me quick. Thanks

rcrichton:
We have decided to hold off on this and hide the channel load column as we can't figure out the deferences yet.

rcrichton:
@BMartinos I think we should close this PR now. Its been too long and I don't think we actually need to put effort into fixing this. What do you think?

BMartinos:
@rcrichton sure that's fine with me. We can re-open the issue again if it becomes problematic but id say close it for now

PR #248, opened 2015-02-26
Removed extra stylesheet include

BMartinos:
Remove stylesheet include as the css has moved to main.css


BMartinos:
Hey @devcritter i have removed the stylesheet include which we no longer use

devcritter:
:+1:

Issue #249, opened 2015-03-02
Re-factor App config setup


BMartinos:
Currently in chrome an error exists where the Api factory gets created before the config variables or retrieved  and set from the HTTP GET. This then creates a application with an undefined server details. Issue noted in this previous user issue: https://github.com/jembi/openhim-console/issues/220

PR #250, opened 2015-03-02
Chrome config bug

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-console/issues/249

Had to re-factor the way the config script is loaded. In chrome angular gets to the Api factory creation faster than in other browsers and by that time the config hasnt been set at certain times. 

Solution: Moved http request to get config details to be the first code to execute and upon receiving a result manually boostrapping the application to ensure config variables exist


BMartinos:
Hey @rcrichton Do you mind just reviewing this PR for me when you get a chance. Thanks

app/scripts/app.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-angular
+var app = angular
   .module('openhimWebui2App', [
```
rcrichton: Maybe we should take this opurtunity to rename the module, its was pretty badly named (by me :) ) Maybe we can make this openhimConsoleApp?

rcrichton:
Great, nicely done @BMartinos looks good to me. One comment above, but nothing to do with your code. :+1:

BMartinos:
Hey @rcrichton i have made that module naming change and then a small change on app.js. I forgot i removed the title variable creation so just added it back in

BMartinos:
Hey @rcrichton i have corrected the broken tests. I also removed some files which doesn't seem to be used or will be used.

rcrichton:
Thanks @BMartinos this looks good :+1:

PR #252, opened 2015-03-05
OHM-64: First version of ATNA UI

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-console/issues/224

Note: I also updated the transaction list view/controller to re-use the 'Load More' button for the audit list


BMartinos:
Hey @rcrichton do you mind just reviewing this PR for me when you get a chance. The ATNA API has been approved and merged into master. 

P.S. I am still looking at adding filter options to the URL but this might be a bit more complicated with the complex filters

rcrichton:
@BMartinos one thing to look at is when I click apply filters after not changing anything I expect it to update the list to show new audit but it doesn't unless I change a filter. I'm still looking through the code, so will have some more comments soon.

app/scripts/controllers/audits.js
```diff
@@ -0,0 +1,410 @@
...
+    /**         Initial page load functions           **/
+    /***************************************************/
+
+
+
+    /*************************************************************/
+    /**         Audits List and Detail view functions           **/
+    /*************************************************************/
+
+    var valeNotEmpty = function(value){
```
rcrichton: typo: vale -> value?

app/scripts/controllers/audits.js
```diff
@@ -0,0 +1,410 @@
...
+      $scope.busyLoadingMore = false;
+    };
+
+    var loadMoreError = function(err){
+      // on error - Hide load more button and show error message
+      jQuery('#loadMoreBtn').hide();
+      Alerting.AlertAddServerMsg(err.status);
+    };
+
+    //location provider - load transaction details
```
rcrichton: I'm guessing this should read load audit details?

app/scripts/services/alerting.js
```diff
@@ -30,6 +30,9 @@ angular.module('openhimConsoleApp')
           case 403:
             alertMsg = 'The request has been forbidden by the server. Please contact the server administrator';
             break;
+          case 404:
+            alertMsg = 'The request could not be found. Please contact the server administrator';
```
rcrichton: Perhaps change this message to: 'the request resource could not be found'.

rcrichton:
Hey @BMartinos looks very nice. Just a few minor comment above. Also, I don't see any unit tests for the controller, that's all that is missing. Have a look at the transactions test for examples.

BMartinos:
Hey @rcrichton i have made those changes as mentioned. Some silly typos :). The tests i made similar to the transactions tests but let me know if i need to add more coverage.

test/spec/controllers/testAudits.js
```diff
@@ -0,0 +1,134 @@
...
+
+  it('should check that the user prefered filters are set', function () {
+    createController();
+    httpBackend.flush();
+
+    // the consoleSession object is setup with user profile in 'login.js'
+    scope.settings.filter.limit.should.equal(10);
+    scope.settings.list.tabview.should.equal('new');
+  });
+
```
rcrichton: I think one other good test to add would be to check that if the filter object is set the correct filters get sent to the httpBackend.

rcrichton:
@BMartinos this looks good to me. Just one comment about an additional test that I think is useful. See above.

BMartinos:
Hey @rcrichton ive added some tests to check the filters object that gets sent to the API. I tried this first with the httpBackend but it does not seem to pick up parameters so i could do an exact match. What i did was set a few filter parameters and run the returnFilters() function to verify the filters are all set correctly as well as the urlParams that get added to the URL

rcrichton:
Cool thanks, looks good to me :+1:

PR #263, opened 2015-03-19
Tasks model updates

### LINKED ISSUE #262: Tasks: Update interface following updated model
The tasks interface needs to be updated to reflect:
* Batch sizes https://github.com/jembi/openhim-core-js/issues/438
* Pausing/Cancelling https://github.com/jembi/openhim-core-js/issues/434

In addition, the current interface uses the `tstatus` field to show the rerun status, instead of the actual `rerunStatus` field. This should be changed since the rerunStatus captures the result of the rerunned transaction.

devcritter:
Closes #262 

### DIFF
--- a/app/scripts/controllers/taskDetails.js
+++ b/app/scripts/controllers/taskDetails.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('openhimConsoleApp')
-  .controller('TaskDetailsCtrl', function ($scope, $modal, $location, $routeParams, Api, Alerting) {
+  .controller('TaskDetailsCtrl', function ($scope, $modal, $location, $routeParams, Api, Alerting, $route) {
 
     /***************************************************/
     /**         Initial page load functions           **/
@@ -56,5 +56,47 @@ angular.module('openhimConsoleApp')
     /**************************************************/
     /**         Task Calculation Functions           **/
     /**************************************************/
+
+    function updateTaskWithStatus(task, status) {
+      var updated = new Api.Tasks();
+      updated._id = task._id;
+      updated.status = status;
+      updated.$update({}, function(){
+        $route.reload();
+      });
+    }
+
+    $scope.pauseTask = function(task){
+      updateTaskWithStatus(task, 'Paused');
+    };
+
+    $scope.resumeTask = function(task){
+      updateTaskWithStatus(task, 'Queued');
+    };
+
+    $scope.cancelTask = function(task){
+      var cancelObject = {
+        title: 'Cancel Task',
+        button: 'Yes',
+        message: 'Are you sure you want to cancel this task?'
+      };
+
+      var modalInstance = $modal.open({
+        templateUrl: 'views/confirmModal.html',
+        controller: 'ConfirmModalCtrl',
+        resolve: {
+          confirmObject: function () {
+            return cancelObject;
+          }
+        }
+      });
+
+      modalInstance.result.then(function () {
+        // cancel confirmed
+        updateTaskWithStatus(task, 'Cancelled');
+      }, function () {
+        // cancel cancelled
+      });
+    };
 
   });
--- a/app/views/transactionsRerunModal.html
+++ b/app/views/transactionsRerunModal.html
@@ -40,6 +40,11 @@
         <option ng-if="transactionsSelected.length >= 64" value="64">64</option>
       </select>
     </div>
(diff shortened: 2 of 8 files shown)


rcrichton:
Hey @devcritter, I'm getting the following when I try resume a task that was started paused:

`error: Could not update Task by ID {taskId} via the API: MongoError: exception: Mod on _id not allowed`

Could be a problem my side.

devcritter:
Hmmm okay - it looks like we're missing a step to delete the _id for updates in the [api](https://github.com/jembi/openhim-core-js/blob/master/src/api/tasks.coffee#L150). It could be that? Weird that I don't get it on my side, which version of mongo are you using? I'm on 2.6.8

app/views/transactionsRerunModal.html
```diff
@@ -28,7 +28,25 @@ <h4 ng-show="transactionsSelected.length > 1">Your selected transactions have be
 
   </div>
   <div class="modal-footer">
+    <div ng-show="transactionsSelected.length > 1" ng-if="rerunSuccess != true" style="float: left;">
+      <label style="float: left; margin: 10px 10px 0px 0px;" tooltip="The number of transactions that the task processor should run concurrently">Batch size</label>
+      <select style="float: left; width: auto;" class="form-control" ng-model="taskSetup.batchSize">
+        <option value="1">One at a time</option>
+        <option ng-if="transactionsSelected.length >= 2" value="2">2</option>
```
rcrichton: Minor: I wonder if we should make these all append 'at a time' eg. 2 at a time etc.

rcrichton:
@devcritter this look good other than the issues I'm running into.

I'm running: `db version v2.4.9`

app/views/transactionsRerunModal.html
```diff
@@ -28,7 +28,25 @@ <h4 ng-show="transactionsSelected.length > 1">Your selected transactions have be
...
+        <option ng-if="transactionsSelected.length >= 4" value="4">4</option>
+        <option ng-if="transactionsSelected.length >= 8" value="8">8</option>
+        <option ng-if="transactionsSelected.length >= 16" value="16">16</option>
+        <option ng-if="transactionsSelected.length >= 32" value="32">32</option>
+        <option ng-if="transactionsSelected.length >= 64" value="64">64</option>
+      </select>
+    </div>
+    <div ng-show="transactionsSelected.length > 0" ng-if="rerunSuccess != true" style="float: left;">
+      <input style="float: left; margin: 10px 10px 0px 0px" type="checkbox" ng-model="taskSetup.paused">
+        <label style="float: left; margin: 10px 10px 0px 0px;" tooltip="Do not start processing the task immediately">Paused</label>
```
rcrichton: This paused checkbox just need to be spaced a bit further from the dropdown.

rcrichton:
I do think it could be due to a _id being passed in the update but this should be the same surely. I don't think you are editing it.

app/scripts/controllers/taskDetails.js
```diff
@@ -57,4 +57,46 @@ angular.module('openhimConsoleApp')
     /**         Task Calculation Functions           **/
     /**************************************************/
 
-  });
\ No newline at end of file
+    function updateTaskWithStatus(task, status) {
+      var updated = new Api.Tasks();
+      updated._id = task._id;
```
rcrichton: Maybe here you can just edit the task object and call `.$update` on it? Is there a need to make a new one?

devcritter:
@rcrichton could perhaps try [this](https://github.com/jembi/openhim-console/pull/263) patch to see if it helps?

app/scripts/controllers/taskDetails.js
devcritter: Yeah, I was trying to avoid that, else it sends through all the fields for the task, like remainingTransactions etc.

rcrichton: Ah ok.

rcrichton:
Thanks that worked.

devcritter:
Aah great. I wonder if they haven't changed that in 2.6? Because on mine updates work fine if you've got the _id in the data object. Haven't found a reference for it yet. But anyway.

I've made the other changes.

rcrichton:
Yeah, I wonder. Its probably best to support older version of mongo anyway.

Cool, looks good :+1:

devcritter:
Yeah, agree

PR #265, opened 2015-03-20
Alert groups w channels

debonair:
@rcrichton please review also review the associtated core pull request


rcrichton:
Looks good @debonair, just one thing. I get an error message like this: `Could not delete the group because it is associtated with the following channels: polling test ,CSD2`

The comma should be before the space :) Once that's done :+1:

PR #271, opened 2015-03-26
Tasks based on filters

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-core-js/issues/436

This branch works with changes made to the transactions API ( currently in branch 'tasks-based-on-filters' )

### DIFF
--- a/app/scripts/controllers/transactions.js
+++ b/app/scripts/controllers/transactions.js
@@ -12,6 +12,8 @@ angular.module('openhimConsoleApp')
 
     // set default limit
     var defaultLimit = 20;
+    var defaultTabView = 'same';
+    $scope.defaultBulkRerunLimit = 10000;
     $scope.loadMoreBtn = false;
 
     // filters collapsed by default
@@ -27,7 +29,7 @@ angular.module('openhimConsoleApp')
     // default settings
     $scope.settings = {};
     $scope.settings.list = {};
-    $scope.settings.list.tabview = 'same';
+    $scope.settings.list.tabview = defaultTabView;
     $scope.settings.filter = {};
     $scope.settings.filter.limit = defaultLimit;
 
@@ -411,7 +413,7 @@ angular.module('openhimConsoleApp')
       }
 
       // Add filters to url
-      // set filter limit only if url parameter set and NOT a bulkrerun (limit not used for bulkrerun so not needed in url)
+      // set filter limit only if url parameter set and NOT a bulkrerun (using default defaultBulkRerunLimit)
       if ( $scope.settings.filter.limit && optionalParam !== 'bulkrerun' ){ $location.search( 'limit', $scope.settings.filter.limit ); }
       if ( $scope.settings.filter.startDate ){ $location.search( 'startDate', $scope.settings.filter.startDate ); }
       if ( $scope.settings.filter.endDate ){ $location.search( 'endDate', $scope.settings.filter.endDate ); }
@@ -474,8 +476,12 @@ angular.module('openhimConsoleApp')
           // add filterRepresentation to only return bulkrerun properties
           returnFilters.filterRepresentation = 'bulkrerun';
 
+          // set filter limit to default defaultBulkRerunLimit
+          returnFilters.filterLimit = $scope.defaultBulkRerunLimit;
+
+
           // remove filter limit and page because all records need to be returned
-          delete returnFilters.filterLimit;
+          // delete returnFilters.filterLimit;
           delete returnFilters.filterPage;
           
           // set bulkrerunActive true to show rerun information
@@ -557,7 +563,7 @@ angular.module('openhimConsoleApp')
       $scope.settings.filter.limit = defaultLimit;
       $scope.settings.filter.startDate = '';
       $scope.settings.filter.endDate = '';
-      $scope.settings.list.tabview = 'same';
+      $scope.settings.list.tabview = defaultTabView;
       $scope.filters.transaction.status = '';
       $scope.filters.transaction.channel = '';
 
--- a/app/scripts/controllers/transactions.js
+++ b/app/scripts/controllers/transactions.js
@@ -479,9 +479,7 @@ angular.module('openhimConsoleApp')
           // set filter limit to default defaultBulkRerunLimit
           returnFilters.filterLimit = $scope.defaultBulkRerunLimit;
 
-
-          // remove filter limit and page because all records need to be returned
-          // delete returnFilters.filterLimit;
+          // remove filterPage because no pagination needed
           delete returnFilters.filterPage;
           
           // set bulkrerunActive true to show rerun information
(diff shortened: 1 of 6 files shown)


BMartinos:
Hey @rcrichton  Do you mind just review this for me when you get a chance. Thanks

This PR depends on the changes made in core (tasks-based-on-filters branch)

app/scripts/controllers/transactions.js
```diff
@@ -26,7 +26,7 @@ angular.module('openhimConsoleApp')
     $scope.settings.list = {};
     $scope.settings.list.tabview = 'same';
     $scope.settings.filter = {};
-    $scope.settings.filter.limit = 10;
+    $scope.settings.filter.limit = 100;
```
rcrichton: Can't we change this back to 10 by default. Any reason why it needs to be changed?

app/views/partials/transaction-filter-settings.html
```diff
@@ -223,6 +223,12 @@
...
   <div class="transactionsFilterRefreshDiv">
-    <button class="btn btn-primary" ng-click="applyFiltersToUrl();"><i class="glyphicon glyphicon-refresh"></i> Apply Filters</button>
-    <button class="btn btn-primary" ng-click="clearFilters();"><i class="glyphicon glyphicon-remove"></i> Clear Filters</button>
+    <div style="float: right; margin-bottom: 5px;">
+      <button class="btn btn-primary" ng-click="applyFiltersToUrl();"><i class="glyphicon glyphicon-refresh"></i> Apply Filters</button>
+      <button class="btn btn-primary" ng-click="clearFilters();"><i class="glyphicon glyphicon-remove"></i> Clear Filters</button>
+    </div>
+
+    <div style="">
+      <button class="btn btn-primary" ng-click="applyFiltersToUrl('bulkrerun');" style="float: right;"><i class="glyphicon glyphicon-repeat"></i> Apply filters for Bulk Rerun ( Find all Transactions )</button>
```
rcrichton: Could we change the text here to read 'Re-run all transaction that match current filters'.

app/views/transactions.html
```diff
@@ -34,7 +34,21 @@
             </tabset>            
           </div>
 
-          <div id="no-more-tables">
+          <!-- Restart the server message -->
+          <alert ng-if="bulkRerunActive" class="alert ng-isolate-scope alert-danger" style="font-weight: bold; text-align: center; font-size: 16px;">
+            <div style="margin-bottom: 10px;">
+              <h2 style="margin: 5px;">You have opted to do a Bulk Rerun!</h2>
+              <h3 style="margin: 20px;">Your filters returned a total of <strong><u>{{transactions.length}}</u></strong> transactions</h3>
```
rcrichton: Could you add '.. to be re-run' to the end of the text.

app/views/transactions.html
```diff
@@ -34,7 +34,21 @@
...
+              <h4>Click '<strong>Continue...</strong>' to setup you rerun procedure <br />
+              or '<strong>Clear Everything!</strong>' if you do not wish to proceed.</h4>
```
rcrichton: Perhaps remove the above two lines. I don't think they are needed.

app/views/transactions.html
```diff
@@ -34,7 +34,21 @@
...
+            </div>
+            <button class="btn btn-danger btn-lg" ng-click="bulkRerunContinue()" style="margin: 0px 5px;">Continue...</button>
+            <button class="btn btn-danger btn-lg" ng-click="clearFilters()" style="margin: 0px 5px;">Clear Everything!</button>
```
rcrichton: Could you make this say 'cancel' rather than 'clear everything'? I think that's more clear.

rcrichton:
Hey @BMartinos the code looks great. Just a few wording comments above.

app/scripts/controllers/transactions.js
BMartinos: No specific reason for the change. If im not mistaken @devcritter mentioned it might be better to show 100 records by default.  Should this rather default to the min amount available (10)? The user can always set their own default in their profile settings

rcrichton: I think 10 is a good default number as otherwise users may not know see the buttons at the end of the list.

devcritter: I still think 10 is too little, but yeah, that's a good a point about seeing the buttons at the end, so yeah, it's probably good to stick to it then. I'll just have to change my user settings :)

app/views/transactions.html
BMartinos: After having having a chat with @devcritter he mentioned it might be good to have a message display if no date range was selected to filter by. If no date range is supplied then all transactions will be returned. 

"No date range has been supplied, Querying all transactions in the defined filters"?

app/scripts/controllers/transactions.js
rcrichton: Maybe 15 or 20?

devcritter: 20 could be good? Unlike 100 it wouldn't take much to scroll down to notice the buttons at the end. Maybe we could also add a tooltip to the limit label at the top saying something like "tip: you can change your default limit in your user profile"?

app/views/transactions.html
rcrichton: Sounds good to me. Maybe just with a slight wording change:

"No date range has been supplied, querying all transactions with the defined filters."

app/scripts/controllers/transactions.js
rcrichton: 20 is fine. The tool tip is a good idea.

BMartinos:
Hey @rcrichton Do you mind reviewing this PR for me when you get a chance. The Travis Builds are still failing however

rcrichton:
Hey @BMartinos This is looking good, but I have a few comments:

* It seems with a massive amount of transactions in the database, the query for them can crash the process due to: `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory` - I'm not too sure what we can do about this, but it seems to take down the whole API until a server restart. Perhaps we should impose a practical limit of like 10 000 transactions to prevent this.
* When loading a large list of transaction the UI display 'Your filters returned a total of 0 transactions that can be re-run' before updating with the correct number. Perhaps we should display a loading icon while the query is being made?

@devcritter it would be good to get your thoughts on this.

BMartinos:
Good point i didn't consider it crashing mongo querying so many records. Having a limit on is a great idea. Should we just set a base limit of say 10 000 or allow the user to choose between a range ( 2000, 4000, 6000, 8000, 10000 ). If no selected then default to 10 000.

Will add in the loading icon till result is returned

rcrichton:
Mongo is actually fine with it, its just the node process that can't cope. It trying to build up this massive JSON object to return to the console and it runs out of memory.

I'd say we just set this to 10 000 behind the scenes and then if a query returns 10 000 results just state that you can't create a bulk re-run with more than 10 000 results, and ask if they want to continue with the first 10 000 transactions matched.

BMartinos:
Hey @rcrichton I have added a default limit of 10 000 transactions to the bulk rerun. Mind just reviewing this for me again. Thanks

rcrichton:
Thanks @BMartinos this is great. :+1:

PR #279, opened 2015-03-30
OHM-138: Added referring URL redirect on login

BMartinos:
This PR relates to issue: #268

### DIFF
--- a/app/config/default.json
+++ b/app/config/default.json
@@ -1,6 +1,6 @@
 {
   "protocol": "https",
-  "host": "localhost",
+  "host": "openhim-preprod.jembi.org",
   "port": 8080,
   "title": "Admin Console",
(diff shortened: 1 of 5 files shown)


BMartinos:
Hey @devcritter Would you mind reviewing this PR for me when you have a chance. One thing that might change is to show the redirectURL as a parameter on the login screen. This will however just be for aesthetics purposes as the actual redirect url is stored in a scope variable

@rcrichton maybe you can also give some feedback regarding the login parameter.

rcrichton:
Hey @BMartinos this works great. I don't think you need to show the redirectURL on the login screen. As it works now is good.

devcritter:
:+1: Yup, works great. Agree that it's necessary to show the redirect url.

PR #284, opened 2015-04-01
hardened create cert forms and added different buttons

debonair:
@rcrichton please review


rcrichton:
Thanks @debonair this looks good to me :+1:

PR #291, opened 2015-04-13
added optional passphrase option

debonair:
@BMartinos @rcrichton please take a look

### DIFF
--- a/app/scripts/controllers/certificates.js
+++ b/app/scripts/controllers/certificates.js
@@ -209,10 +209,25 @@ angular.module('openhimConsoleApp')
       $scope.certificateObject = new Api.Keystore();
       $scope.certificateObject.passphrase = $scope.serverPassphrase;
       $scope.certificateObject.$save({ type: 'passphrase' }, function(){
-        $scope.uploadSuccess('serverKey', '');
+        $scope.passphraseSuccess('serverKey', '');
       }, function(err){
-        $scope.uploadFail(err, 'serverKey', '');
+        $scope.passphraseFail(err, 'serverKey', '');
       });
+    };
+
+    $scope.passphraseSuccess = function(){
+      $scope.importSuccess++;
+      $scope.serverRestartRequired = true;
+      $scope.resetCertificates();
+      $scope.goToTop();
+    };
+
+    $scope.passphraseFail = function(location){
+      Alerting.AlertAddMsg(location, 'danger', 'The passphrase does not match the key');
+      $scope.importSuccess++;
+      $scope.serverRestartRequired = true;
+      $scope.resetCertificates();
+      $scope.goToTop();
     };
 
 
--- a/app/scripts/controllers/certificates.js
+++ b/app/scripts/controllers/certificates.js
@@ -217,6 +217,7 @@ angular.module('openhimConsoleApp')
       $scope.importSuccess++;
       $scope.serverRestartRequired = true;
       $scope.resetCertificates();
+      $scope.serverPassphrase = null;
       $scope.goToTop();
     };
 
--- a/app/scripts/controllers/certificates.js
+++ b/app/scripts/controllers/certificates.js
@@ -213,7 +213,8 @@ angular.module('openhimConsoleApp')
       });
     };
 
-    $scope.passphraseSuccess = function(){
+    $scope.passphraseSuccess = function(location){
+      Alerting.AlertAddMsg(location, 'success', 'Passphrase submitted');
       $scope.importSuccess++;
       $scope.serverRestartRequired = true;
       $scope.resetCertificates();
--- a/app/scripts/controllers/certificates.js
+++ b/app/scripts/controllers/certificates.js
@@ -329,8 +329,4 @@ angular.module('openhimConsoleApp')
       });
     };
 
-    $scope.addServerPassphrase = function() {
-
-    };
-
   });
--- a/app/scripts/controllers/certificates.js
+++ b/app/scripts/controllers/certificates.js
@@ -205,6 +205,16 @@ angular.module('openhimConsoleApp')
       }
     };
 
+    $scope.addPassphrase = function () {
+      $scope.certificateObject = new Api.Keystore();
+      $scope.certificateObject.passphrase = $scope.serverPassphrase;
+      $scope.certificateObject.$save({ type: 'passphrase' }, function(){
+        $scope.uploadSuccess('serverKey', '');
+      }, function(err){
+        $scope.uploadFail(err, 'serverKey', '');
+      });
+    };
+
 
     /****************************************/
     /**         Import Functions           **/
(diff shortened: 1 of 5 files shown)


app/views/certificates.html
```diff
@@ -122,6 +122,8 @@
 
               <!-- Server Certificate Alerts -->
               <alert ng-repeat="alert in alerts.serverKey" type="alert.type" style="margin: 20px 0px 0px 0px;">{{alert.msg}}</alert>
+              <h4><strong>Enter a passphrase (Optional)</strong></h4>
+              <input class="form-control" placeholder="Enter a passphrase here if the key requires one." type="text" ng-model="serverPassphrase" />
```
rcrichton: I think that this should be a password input type. So the password into in plain text. Also, it may be better if it has a fixed width so it doesn't become so long.

app/scripts/controllers/certificates.js
```diff
@@ -108,6 +108,8 @@ angular.module('openhimConsoleApp')
       switch ( $scope.uploadType ){
         case 'serverCert':
           $scope.certificateObject.cert = data;
+          $scope.certificateObject.passphrase = $scope.serverPassphrase;
```
rcrichton: Do we need this on the serverCert? Wouldn't we only need it on the key?

rcrichton:
Hey @debonair some quick comment above.

app/scripts/controllers/certificates.js
debonair: This would be necessary for both, for usability's sake if you modify either of them you want the passphrase field sent in the request as well, I was avoiding having a seperate button

BMartinos: I don't think this would be needed for the serverCert. Although nice to have that functionality in but based on the text input location it seems like it can only be supplied by the serverKey.

app/views/certificates.html
```diff
@@ -60,8 +60,8 @@
                   </td>
                 </tr>
                 <tr>
-                  <td width="120">Comman Name: </td>
```
BMartinos: maybe have a toggle option to not show the input field by default. something like

<a data-toggle="collapse" data-target="#passphrase">Supply a passphrase</a>
<div class="collapse" id="passphrase">
input box to show on toggle
</div>

BMartinos:
Maybe just clear the $scope.serverPassphrase variable in the  $scope.$watch('serverKey' function so user can start fresh after uploading

BMartinos:
Hey @debonair. Just some comments on the passphrase textbox, 

* Should validate if textbox is empty before trying to submit. Currently empty value can be submitted. 
* The upload message should rather reference something regarding the passphrase and not 'Newly Uploaded File'. Maybe something like "Passphrase successfully saved"
* The textbox should be cleared once the save button is clicked. 

The process of how the user saves the passphrase works well for me and i dont think there is much more that can be done to better streamline the save.

app/scripts/controllers/certificatesmodal.js
```diff
@@ -66,9 +66,11 @@ angular.module('openhimConsoleApp')
       }
 
       // country validation
-      if(($scope.cert.country.length === 2) === false){
-        $scope.ngError.country = true;
-        $scope.ngError.hasErrors = true;
+      if ($scope.cert.country) {
+        if(($scope.cert.country.length === 2) === false){
```
rcrichton: You can use `$scope.cert.country.length !== 2` here rather

app/views/certificates.html
```diff
@@ -122,6 +122,9 @@
 
               <!-- Server Certificate Alerts -->
               <alert ng-repeat="alert in alerts.serverKey" type="alert.type" style="margin: 20px 0px 0px 0px;">{{alert.msg}}</alert>
+              <h4><strong>Enter a passphrase (Optional)</strong></h4>
```
rcrichton: Maybe instead of 'Optional' we should say 'If required'

app/scripts/controllers/certificates.js
```diff
@@ -201,6 +203,33 @@ angular.module('openhimConsoleApp')
...
+      $scope.goToTop();
+    };
+
+    $scope.passphraseFail = function(location){
+      Alerting.AlertAddMsg(location, 'danger', 'The passphrase does not match the key');
+      $scope.importSuccess++;
+      $scope.serverRestartRequired = true;
+      $scope.resetCertificates();
+      $scope.goToTop();
+    };
```
rcrichton: Please could you add some simple unit test for the above 3 functions.

app/scripts/controllers/certificates.js
```diff
@@ -201,6 +203,33 @@ angular.module('openhimConsoleApp')
...
+      Alerting.AlertAddMsg(location, 'success', 'Passphrase submitted');
+      $scope.importSuccess++;
+      $scope.serverRestartRequired = true;
+      $scope.resetCertificates();
+      $scope.serverPassphrase = null;
+      $scope.goToTop();
+    };
+
+    $scope.passphraseFail = function(location){
+      Alerting.AlertAddMsg(location, 'danger', 'The passphrase does not match the key');
```
rcrichton: I don't seen to ever get this alert even when I enter a wrong passphrase.

app/views/certificates.html
```diff
@@ -122,6 +122,9 @@
...
+              <input class="form-control pull-left" placeholder="Enter a passphrase here if the key requires one." type="password" ng-model="serverPassphrase" style="width: 70%;" />
```
rcrichton: I would be great it we could show a symbol next to this text box to show if a password has already been submitted. See the client modal basic auth password box, Martin has done something similar there.

rcrichton:
Hey @debonair, looking good. A few comments above. Please prioritise this work so we can release later today.

app/views/certificates.html
debonair: Adding a symbol will not be possible without making changes to the core. I would need to return the whole keystore to get the passphrase bit

rcrichton: Ok, that is fine then. We can leave it as is for now.

app/scripts/controllers/certificates.js
```diff
@@ -201,6 +203,37 @@ angular.module('openhimConsoleApp')
...
+        $scope.passphraseSuccess('serverKey', '');
+      }, function(err){
+        $scope.passphraseFail(err, 'serverKey', '');
+      });
+    };
+
+    $scope.passphraseSuccess = function(location){
+      Api.Keystore.get({ type: 'validity' }, function(result){
+        $scope.certValidity = result;
+        Alerting.AlertAddMsg(location, 'success', 'Passphrase submitted');
```
rcrichton: Maybe instead of saying 'passphrase submitted' here we should say 'Passphrase matches supplied key'. This will work better with the fail case.

test/spec/controllers/certificates.js
```diff
@@ -21,9 +21,9 @@ describe('Controller: CertificatesCtrl', function () {
 
     httpBackend = $httpBackend;
 
-    $httpBackend.when('GET', new RegExp('.*/keystore/validity')).respond({'valid': true});
+//    httpBackend.when('GET', new RegExp('.*/keystore/validity')).respond({'valid': true});
```
rcrichton: Remove commented out code.

rcrichton:
Hey @debonair just a super minor comment above otherwise :+1: This works well with the core fix.

Issue #290, opened 2015-04-09
Contact Groups -> Contact List


devcritter:
Despite #114 our terminology can still be a bit confusing with regards to users. There are permissions groups on a channel, but then there are also contact groups for alerts and reports, which is something different.

To avoid confusion, should we rename Contact Groups into Contact Lists?

It's probably not a good idea to update the schema at this stage, so this would be a cosmetic change.

BMartinos:
I like the proposed idea of changing the label names from 'Contact Groups' to 'Contact Lists'. This is a quick easy fix to overcome this confusion.

rcrichton:
Yeah, sounds like a good idea to me as well. We can't change the schema unless we bump the major version so a cosmetic change sounds good.

PR #292, opened 2015-04-13
Contact group label change

BMartinos:
This PR relates to issue: #290


BMartinos:
Hey @devcritter Ive updated the contact groups label where i could spot them. Not sure if ive missed any locations. It would be a good idea to update the function and file names as well as this might become confusing. Should i do these first?

devcritter:
Great thanks, looks good. Just two more places I could see:

The *group* column should be *List*
screen shot 2015-04-17 at 12 26 38
and when adding/editing a group, the *Group Name* should be *List Name*
screen shot 2015-04-17 at 12 26 57

Feel free to merge after those changes :+1:

PR #301, opened 2015-04-28
Added check to see if session data exists

devcritter:
@rcrichton a quick one - I ran into a bug on a clean install.


rcrichton:
:+1:

PR #314, opened 2015-07-09
Issue #307: Only show headers if available.

### LINKED ISSUE #307: Remove headers from transaction if none supplied
Currently the headers details always display regardless if there are any. this must be removed if headers are not set

devcritter:
Only show response section if response obj available.

Closes #307 

@BMartinos a really quick one. Was this what you had in mind for the ticket?


BMartinos:
@devcritter yip that is what i had in mind regarding the headers being hidden if none exist. This looks good to me. Might be useful to do this on the routes/orchestrations views as well.
https://github.com/jembi/openhim-console/blob/master/app/views/transactionsAddReqResModal.html#L37

devcritter:
Great thanks @BMartinos, added that in

BMartinos:
Great that looks good to me :+1:

PR #315, opened 2015-07-10
Autorefresh

### LINKED ISSUE #299: Tasks Auto Refresh
Implement functionality to automatically update tasks.

This was originally implemented in UI changes #289 but we noticed a logic issue that if clicking on 'load more' and it refreshes it brings back the default limit and not the extra tasks

e.g. Load more to view 20 tasks -> refresh happens -> 10 records returned

### LINKED ISSUE #305: Transactions log auto refresh
Similar to #299 

it would be really cool if the transactions page could also auto-refresh

devcritter:
Closes #299 
Closes #305 

### DIFF
--- a/app/views/transactions.html
+++ b/app/views/transactions.html
@@ -98,7 +98,7 @@
                 </tr>
               </thead>
               <tbody>
-                <tr class="table-list" ng-repeat="transaction in transactions" ng-style="baseIndex + $index < 0 && {'color': '#036633', 'font-weight': 'bold'}" ng-click='viewTransactionDetails("transactions/" + transaction._id, $event)'>
+                <tr class="table-list" ng-repeat="transaction in transactions" ng-class="{transactionsNew: baseIndex + $index < 0}" ng-click='viewTransactionDetails("transactions/" + transaction._id, $event)'>
                   <td class="transactionsCheckboxTD">
                     <label>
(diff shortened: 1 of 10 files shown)


app/views/transactions.html
```diff
@@ -88,27 +98,34 @@ <h3 style="margin: 20px;" ng-if="transactionsSelected.length < defaultBulkRerunL
                 </tr>
               </thead>
               <tbody>
-                <tr class="table-list" ng-repeat="transaction in transactions" ng-click='viewTransactionDetails("transactions/" + transaction._id, $event)'>
+                <tr class="table-list" ng-repeat="transaction in transactions" ng-style="baseIndex + $index < 0 && {'color': '#036633', 'font-weight': 'bold'}" ng-click='viewTransactionDetails("transactions/" + transaction._id, $event)'>
```
rcrichton: Maybe rather apply a css class rather than applying the css inline. I'm not quite sure what this is doing here either, mind explaining a bit.

rcrichton:
Looks good, maybe a few tests for the newly added functions would be useful.

app/views/transactions.html
devcritter: Sure, I'll try switching to a class.

So basically `baseIndex` decrements every time a new transaction comes in, helping to keep to track of which ones were loaded when the page was loaded. If a transaction has `baseIndex+$index` equal or greater than zero, we know it's one of the old transactions; below zero it's new. The old transactions can then keep their original index number in the display. This little ng-style bit applies styling to new transactions that come in.

devcritter:
@rcrichton addressed the comments if you could take another look?

rcrichton:
Cool, looks good to me :+1:

PR #316, opened 2015-07-20
Removes CDNify

rcrichton:
It was freezing the build and I don't think we actually want this in the build process.

@devcritter could you have a quick look at this and see if you agree.


devcritter:
:+1:

PR #327, opened 2015-09-21
OHM-194: Added optional client info

devcritter:
Depends on https://github.com/jembi/openhim-core-js/issues/535

### DIFF
--- a/app/views/clients.html
+++ b/app/views/clients.html
@@ -22,6 +22,10 @@
                   <th width="50">#</th>
                   <th>ID</th>
                   <th>Name</th>
+                  <th>Organization</th>
+                  <th>Software Name</th>
+                  <th>Description</th>
+                  <th>Contact Person</th>
                   <th>Domain</th>
                   <th>Roles</th>
                   <th width="80"></th>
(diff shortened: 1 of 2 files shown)


app/views/clients.html
```diff
@@ -32,6 +36,10 @@
                   <td data-title="#{{ $index +1 }}"><span id="counter">{{ $index +1 }}</span></td>
                   <td data-title="ID"><a href="" ng-click="editClient(client)">{{ client.clientID }}</a></td>
                   <td data-title="Name"><a href="" ng-click="editClient(client)">{{ client.name }}</a></td>
+                  <td data-title="Organization">{{ client.organization }}</td>
+                  <td data-title="SoftwareName">{{ client.softwareName }}</td>
+                  <td data-title="Description">{{ client.description }}</td>
+                  <td data-title="Contact">{{ client.contactPerson }} ({{client.contactPersonEmail}})</td>
```
rcrichton: For this field the brackets still display if the email is empty. I think it would be better to not display the brackets when there is no email address.

rcrichton:
@devcritter Looks good to me, just one comment above.

devcritter:
@rcrichton good idea, added checks. Also added the location field.

rcrichton:
:+1:

PR #328, opened 2015-09-23
Mediator config

### LINKED ISSUE #319: Update mediator page to allow for mediator configuration
See design here: https://docs.google.com/document/d/1y-R5PxQBktoN2qdeNkGMGsgKAU7_VNGF_tkXbdIbVjk/edit#heading=h.9i68ar6rcld1

devcritter:
Closes #319 

@rcrichton if you could please review?

here are some screenshots:
screen shot 2015-09-23 at 09 03 08
screen shot 2015-09-23 at 09 03 47
screen shot 2015-09-23 at 09 03 56

### DIFF
--- a/app/scripts/services/mediatorDisplay.js
+++ b/app/scripts/services/mediatorDisplay.js
@@ -1,0 +1,39 @@
+'use strict';
+/* global moment: false */
+
+angular.module('openhimConsoleApp')
+  .factory('MediatorDisplay', function (config) {
+
+    var _formatMediator = function(mediator){
+      var secondsDiffNow = function(){
+        return Math.abs(Date.now()-moment(mediator._lastHeartbeat)) / 1000;
+      };
+
+      if (!mediator._lastHeartbeat) {
+        mediator.lastHeartbeatStatus = 'never';
+      } else {
+        if (secondsDiffNow() < config.mediatorLastHeartbeatWarningSeconds) {
+          mediator.lastHeartbeatStatus = 'success';
+        } else if (secondsDiffNow() < config.mediatorLastHeartbeatDangerSeconds) {
+          mediator.lastHeartbeatStatus = 'warning';
+        } else {
+          mediator.lastHeartbeatStatus = 'danger';
+        }
+        mediator.lastHeartbeatDisplay = moment(mediator._lastHeartbeat).fromNow();
+      }
+
+      if (mediator._uptime) {
+        //generate human-friendly display string, e.g. 4 days
+        mediator.uptimeDisplay = moment().subtract(mediator._uptime, 'seconds').fromNow(true);
+      }
+    };
+
+    return {
+      formatMediator: _formatMediator,
+
+      formatMediators: function (mediators) {
+        angular.forEach(mediators, _formatMediator);
+      }
+    };
+
+  });
--- a/app/scripts/services/mediatorDisplay.js
+++ b/app/scripts/services/mediatorDisplay.js
@@ -16,8 +16,10 @@ angular.module('openhimConsoleApp')
           mediator.lastHeartbeatStatus = 'success';
         } else if (secondsDiffNow() < config.mediatorLastHeartbeatDangerSeconds) {
           mediator.lastHeartbeatStatus = 'warning';
+          mediator.lastHeartbeatDisplayExplain = 'No heartbeats received in over ' + config.mediatorLastHeartbeatWarningSeconds + ' s';
         } else {
           mediator.lastHeartbeatStatus = 'danger';
+          mediator.lastHeartbeatDisplayExplain = 'No heartbeats received in over ' + config.mediatorLastHeartbeatDangerSeconds + ' s';
         }
         mediator.lastHeartbeatDisplay = moment(mediator._lastHeartbeat).fromNow();
       }
--- a/app/scripts/services/mediatorDisplay.js
+++ b/app/scripts/services/mediatorDisplay.js
@@ -11,6 +11,7 @@ angular.module('openhimConsoleApp')
 
       if (!mediator._lastHeartbeat) {
         mediator.lastHeartbeatStatus = 'never';
+        mediator.lastHeartbeatDisplayExplain = 'No heartbeats have ever been received from this mediator. Perhaps it doesn\'t support heartbeats.';
       } else {
         if (secondsDiffNow() < config.mediatorLastHeartbeatWarningSeconds) {
           mediator.lastHeartbeatStatus = 'success';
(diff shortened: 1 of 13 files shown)


rcrichton:
Hey @devcritter this is looking cool!

I have a few comments/suggestions so far from looking at the screenshots:

1. I think it would look better if the last heartbeat was displayed with a green|yellow|red dot in front of it rather than colouring the text background. Also, tooltips explaining what each colour means would be nice.
2. I think we should have a config icon on the list page (before the delete x) as well as the details page for mediators. On the details page we should probably have a bigger button that say config. I also think the icon should be a cog rather (`glyphicon glyphicon-cog`).
3. I like that the uptime says '6mins' it would be great if we could do the same with the heartbeat. Something like: '20s ago'
4. (optional idea) When a text box becomes more than one line it would be great to change the display to a text area. I'm thinking of when providing blocks of text, like certificates, this may be useful.

app/views/mediatorConfigModal.html
```diff
@@ -0,0 +1,48 @@
...
+    <fieldset>
+
+      <div class="form-group-2columns" ng-repeat="def in mediator.configDefs">
+        <div class="form-group" style="float: left;">
+          <label>{{def.param}}</label>
+        </div>
+        <div class="form-group" style="float: right;">
+
+          <div ng-if="def.type === 'string'">
+            <input class="form-control" type="text" ng-model="mediator.config[def.param]" ng-disabled="update">
```
rcrichton: So currently in core we have a `def.default` property (which isn't be used here) and the mediator can supply it own default config in the `config` property. I now think this is duplicative. Perhaps we should remove the `def.default` property from core and just use the defaults provided in the `config` property. What do you think?

devcritter: Hmmm, yeah agree

devcritter:
Hey @rcrichton tried the icon idea, but it doesn't really work too well:
screen shot 2015-09-23 at 15 39 41

The page has quite a bit of text on it, so the icon might disappear a bit in the noise (a good point raised by @BMartinos ), esp. if you have a lot of mediators.

Text seems clearer:
screen shot 2015-09-23 at 15 47 13

what do you think?

rcrichton:
@devcritter now that its not a timestamp and has clean text about the last time a heartbeat was received I actually like it much better with the background colour. I'd now vote to keep it like that (sorry :) )

devcritter:
@rcrichton ok cool, made the changes. I moved the cog icon to the left on the mediator details page.

Regarding the text box: for now I just made it a text area with 1 row, the user can always drag it to a bigger size if they need it for something like a cert.

One idea with regards could be to have another datatype for config? something like `bigstring` for this use case? then we can keep the simpler inputs for small strings

screen shot 2015-09-23 at 16 32 19
screen shot 2015-09-23 at 16 32 43
(in truth I'm actually just jealous of the BSIS team's pretty picture filled PRs, so I'm just gonna keep putting these in :) )

BMartinos:
lol for the pretty picture'd PR's :). They are very cool though :)

debonair:
Yay! pretty pictures FTW!

rcrichton:
Haha, agreed. The OpenHIM PRs desperately  need more pictures in comparison.

rcrichton:
Hey @devcritter I'm noticed something strange. It seems if a mediator does not register with a `config` property then it is not possible to type any text into the input box on the mediator config page. I think you may just have to define an empty config object if one does not already exist to make the problem go away.

rcrichton:
@devcritter I think we should probably go back to using text boxes rather than text areas. I think your idea of a 'bigstring' is a better one. We can add that in the future.

Another thing that I think we can add are some tooltips that give the exact heartbeat timestamp or uptime value when you hover over the text.

test/spec/controllers/mediators.js
```diff
@@ -88,4 +97,19 @@ describe('Controller: MediatorsCtrl', function () {
...
+  });
+
+  it('should set the uptimeDisplay field with a human friendly display string', function () {
+    createController();
+    httpBackend.flush();
+
+    scope.mediators[0].uptimeDisplay.should.equal('an hour');
+    scope.mediators[1].uptimeDisplay.should.equal('2 months');
+  });
+});
```
rcrichton: Perhaps some test of the mediator details page would be good? Just to make sure things are added to the config object properly.

app/scripts/services/mediatorDisplay.js
```diff
@@ -0,0 +1,41 @@
...
+angular.module('openhimConsoleApp')
+  .factory('MediatorDisplay', function (config) {
+
+    var _formatMediator = function(mediator){
+      var secondsDiffNow = function(){
+        return Math.abs(Date.now()-moment(mediator._lastHeartbeat)) / 1000;
+      };
+
+      if (!mediator._lastHeartbeat) {
+        mediator.lastHeartbeatStatus = 'never';
```
rcrichton: An explanation of never would be nice. Perhaps something like: 'No heartbeat have ever been received from this mediator. Perhaps it doesn't support heartbeats.'

app/views/mediatorDetails.html
```diff
@@ -37,6 +37,17 @@
               <td>{{mediatorDetails.version}}</td>
             </tr>
             <tr>
+              <td><strong>Last Heartbeat: </strong></td>
+              <td title="{{mediatorDetails.lastHeartbeatDisplay}}">
+                <span ng-if="mediatorDetails.lastHeartbeatStatus === 'never'" class="label label-default">Never</span>
+                <span ng-if="mediatorDetails.lastHeartbeatStatus !== 'never'" class="label label-{{mediatorDetails.lastHeartbeatStatus}}" title="{{mediatorDetails.lastHeartbeatDisplayExplain}}">{{mediatorDetails.lastHeartbeatDisplay}}</span>
```
rcrichton: I think instead of using `title=` we should use the tooltips that we have been using in the transaction log. Just to be more consistent. See https://github.com/jembi/openhim-console/blob/master/app/views/transactions.html#L29

devcritter:
@rcrichton made the changes

rcrichton:
@devcritter I'm still getting issues when I register a mediator with no `config` property.

```
TypeError: Cannot set property 'dhisAsync' of undefined
    at extend.assign (angular.js:10537)
    at $setViewValue (angular.js:17312)
    at ui-bootstrap-tpls.js:366
    at Scope.$eval (angular.js:12412)
    at Scope.$apply (angular.js:12510)
    at HTMLLabelElement.<anonymous> (ui-bootstrap-tpls.js:365)
    at HTMLLabelElement.jQuery.event.dispatch (jquery.js:4409)
    at HTMLLabelElement.elemData.handle (jquery.js:4095)
```

Also, as a more minor point, I think we should include a timestamp in the tooltip for each of the places where tmestamps are made human readable.

devcritter:
Argh, sorry coded but didn't test it originally - it works now

rcrichton:
:+1: Nice, this is pretty cool to play with :)

devcritter:
Yeah :)

PR #332, opened 2015-10-05
Removed seconds tooltip from mediator uptime field

devcritter:
@rcrichton just looking at this again, the seconds tooltip on the uptime field is quite ugly. A value in seconds value beyond a couple of minutes really has no meaning to a user. Not sure what i was thinking putting it in :)


rcrichton:
Yeah, you are probably right. I didn't mind it too much through. I wonder if we replace it with a tooltip that says 'Up since 2015-10-06T06:56:08' would be better. We can pretty easily work out the up since date by subtracting the uptime from the current date.

devcritter:
Like this?

"Up since ..." looked a bit weird if the mediator was down so I opted for "Startup time ..."

rcrichton:
Cool yeah, that looks good :+1:

PR #333, opened 2015-10-06
Config def updates

rcrichton:
Related to issue https://github.com/jembi/openhim-core-js/issues/555

Hey @devcritter could you please review. This PR requires the changes [in this core PR to work](https://github.com/jembi/openhim-core-js/pull/556).

@BMartinos could you also have a look at some of the CSS changes and see if these are ok, as discussed.

This is how it looks atm. I'm not sure on the '?' placement, but not sure where else to put them.

med-config

### DIFF
--- a/app/scripts/controllers/mediatorDetails.js
+++ b/app/scripts/controllers/mediatorDetails.js
@@ -7,15 +7,28 @@ angular.module('openhimConsoleApp')
     /**         Initial page load functions           **/
     /***************************************************/
 
+    var createParamDefMap = function (mediator) {
+      var map = {};
+      Object.keys(mediator.config).map(function (param) {
+        map[param] = mediator.configDefs.filter(function (def) {
+          return def.param === param;
+        })[0];
+      });
+      return map;
+    };
+
     var querySuccess = function(mediatorDetails){
       MediatorDisplay.formatMediator(mediatorDetails);
       $scope.mediatorDetails = mediatorDetails;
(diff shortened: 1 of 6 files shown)


BMartinos:
Hey @rcrichton the CSS changes look good to me. I didnt run the branch, only checked the css and the gif ( very cool btw :) ). Regarding the "?" positioning, i think it looks fine the way it is now but there are quite a few of them. Another option is to have the tooltip on the label itself?

rcrichton:
Thanks Martin, I did think of that. The problem is that only some fields will have a tooltip. So it would be a pretty inconsistent user experience if they don't know which fields have a tooltip before trying to mouse over them.

debonair:
I like the animated gif :) what did you use to make it?

rcrichton:
I had to try be just as cool as you guys on BSIS with all your images ;)

I used byzanz (`$ sudo apt-get install byzanz`), it's a command line tool. You can use something like `$ byzanz-record --x=2311 --y=105 --width=1239 --height=950 -d 25 -c file-queue.gif`. For hints for the parameters values use `$ xwininfo`

devcritter:
Looks good to me :+1: I'm also okay with having the '?' - it also helps let the user know that there is a help description available in the first place.

Also. The gif is cool! :)

BMartinos:
@rcrichton I agree, keeping the icon in indicates to the user there is a tooltip. I would keep it the way you have it.

PR #337, opened 2015-11-03
View server logs

### LINKED ISSUE #334: Displaying the OpenHIM server logs on a page in the console would be useful
Related to https://github.com/jembi/openhim-core-js/issues/558

rcrichton:
Closes #334

@devcritter could you please review this for me?

@BMartinos, you may want to have a look as well if you have space, otherwise don't worry.

server-logs

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -146,7 +146,6 @@
     <script src="bower_components/angular-highlightjs/build/angular-highlightjs.js"></script>
     <script src="bower_components/ng-file-upload/angular-file-upload.js"></script>
     <script src="bower_components/ng-file-upload-shim/angular-file-upload-shim.js"></script>
-    <script src="bower_components/moment-timezone/builds/moment-timezone-with-data-2010-2020.js"></script>
     <script src="bower_components/eonasdan-bootstrap-datetimepicker/build/js/bootstrap-datetimepicker.min.js"></script>
     <script src="bower_components/angular-bootstrap-datetimepicker-directive/angular-bootstrap-datetimepicker-directive.js"></script>
     <!-- endbower -->
--- a/app/index.html
+++ b/app/index.html
@@ -16,6 +16,7 @@
     <link rel="stylesheet" href="bower_components/bootstrap-datepicker/css/datepicker3.css" />
     <link rel="stylesheet" href="bower_components/morris.js/morris.css" />
     <link rel="stylesheet" href="bower_components/highlightjs/styles/default.css" />
+    <link rel="stylesheet" href="bower_components/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.min.css" />
     <!-- endbower -->
     <!-- endbuild -->
     <!-- build:css({.tmp,app}) styles/main.css -->
@@ -79,6 +80,7 @@
                             <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/users')}"><a href="#/users"><i class="glyphicon glyphicon-user"></i> Users</a></li>
                             <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/certificates')}"><a href="#/certificates"><i class="glyphicon glyphicon-certificate"></i> Certificate</a></li>
                             <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/export-import')}"><a href="#/export-import"><i class="glyphicon glyphicon-transfer"></i> Export/Import</a></li>
+                            <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/logs')}"><a href="#/logs"><i class="glyphicon glyphicon-cog"></i> Server Logs</a></li>
                           </ul>
                         </li>
                       </ul>
@@ -144,6 +146,9 @@
     <script src="bower_components/angular-highlightjs/build/angular-highlightjs.js"></script>
     <script src="bower_components/ng-file-upload/angular-file-upload.js"></script>
     <script src="bower_components/ng-file-upload-shim/angular-file-upload-shim.js"></script>
+    <script src="bower_components/moment-timezone/builds/moment-timezone-with-data-2010-2020.js"></script>
+    <script src="bower_components/eonasdan-bootstrap-datetimepicker/build/js/bootstrap-datetimepicker.min.js"></script>
+    <script src="bower_components/angular-bootstrap-datetimepicker-directive/angular-bootstrap-datetimepicker-directive.js"></script>
     <!-- endbower -->
     <script src="bower_components/cryptojslib/rollups/sha512.js"></script>
     <script src="bower_components/cryptojslib/components/enc-base64-min.js"></script>
@@ -185,6 +190,7 @@
     <script src="scripts/controllers/audits.js"></script>
     <script src="scripts/controllers/auditDetails.js"></script>
     <script src="scripts/controllers/auditsContentModal.js"></script>
+    <script src="scripts/controllers/logs.js"></script>
 
     <!--- This is for utility functions -->
     <script src="scripts/util/sidebar.js"></script>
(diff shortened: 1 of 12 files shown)


app/index.html
```diff
@@ -144,6 +146,8 @@
     <script src="bower_components/angular-highlightjs/build/angular-highlightjs.js"></script>
     <script src="bower_components/ng-file-upload/angular-file-upload.js"></script>
     <script src="bower_components/ng-file-upload-shim/angular-file-upload-shim.js"></script>
+    <script src="bower_components/eonasdan-bootstrap-datetimepicker/build/js/bootstrap-datetimepicker.min.js"></script>
```
devcritter: I take you had issues with the existing datepicker? If so, I wonder if we shouldn't swap out the existing ones for the new one so as to make it consistent?

cc @BMartinos

app/scripts/controllers/logs.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+/* global moment: false */
+
+angular.module('openhimConsoleApp')
+  .controller('LogsCtrl', function ($scope, $location, Api) {
+    function formatLog(log) {
+      return new moment(log.timestamp).format('YYYY-MM-D HH:mm:ss.SSS') + ' - ' + log.level + ': [' + log.label + '] ' + log.message + '\n';
```
devcritter: Only one D in the date format

app/index.html
BMartinos: Initially time based filtering wasn't needed for transactions/audits but I agree with @devcritter that keeping it consistent will be better. We might just need to check but I cant see any reason why swapping the old one with the new one shouldn't return the same result. The date that will be sent to the API will just contain extra "time" information

rcrichton: I didn't have issues with the existing datepicker but I needed a date**time**picker instead. So I had to find another one.

app/scripts/controllers/logs.js
```diff
@@ -0,0 +1,96 @@
...
+      var now = new Date();
+      var localParam = {
+        from: new Date(lastFetch).toISOString(),
+        until: now.toISOString(),
+        level: $scope.params.level
+      };
+      lastFetch = now;
+      Api.Logs.query(localParam, function (results) {
+        results.forEach(function (log) {
+          $scope.logs += formatLog(log);
```
devcritter: It might be good to enforce a max size here - maybe keep a line count if it's above 100k then start removing from the front?

app/scripts/controllers/logs.js
```diff
@@ -0,0 +1,96 @@
...
+      // clear existing intervals whenever the location is changed
+      clearInterval(autoUpdateInterval);
+      clearInterval(scrollInterval);
+    });
+
+    $scope.reset = function() {
+      delete $scope.params.from;
+      delete $scope.params.until;
+      $scope.params.level = 'info';
+      $scope.reload();
```
devcritter: Atm when you hit 'clear' on the interface it keeps the existing data, which is a bit annoying when the logs have filled up a lot. It would be good to set $scope.logs to empty here, I guess before calling reload.

BMartinos:
Hey @rcrichton I havent looked to much into the code yet but from looking at the awesome gif :) i think the UI/UX is looking great! Just one thing i noticed is that if you click on a date it automatically adds in the current time ( 13:45 ), Would it be possible to have this as 00:00 and then give the user the choice to specify the time if they need it? 

Otherwise, i think it looks great!

app/index.html
devcritter: They've got different stylings though. Does the new one have an option to use a datepicker (no -time)? Tbh I'm not so crazy about them co-existing due to the inconsistent styling.

app/scripts/controllers/logs.js
```diff
@@ -0,0 +1,96 @@
...
+    var lastFetch;
+    var locParams = $location.search();
+    $scope.params = angular.equals({}, locParams) ? { level: 'info' } : locParams;
+    $scope.logs = '';
+
+    if ($scope.params.from || $scope.params.until) {
+      $scope.autoupdate = false;
+      $scope.tailLogs = false;
+    } else {
+      $scope.autoupdate = true;
```
devcritter: Can you add a control to toggle autoupdate on the interface? It would be useful if you want to study the logs without having it consistently updating. Another idea could be to merge tail and autoupdate? I.e. only autoupdate when tail is true?

devcritter:
@rcrichton looks good so far - just some comments above.

app/scripts/controllers/logs.js
rcrichton: Hmm, it should be clearing the logs when it reloads the controller. Is this consistently not working for you?

app/scripts/controllers/logs.js
rcrichton: The core API is limited to 100 000 lines at a maximum, do you think we should limit here as well?

app/scripts/controllers/logs.js
BMartinos: Not sure if this could be the issue but if you trying to reload the controller by setting the location.search paramaters to the same parameters then the reload doesn't happen. Angular ignores this then as the parameters are the same

app/index.html
rcrichton: Hmm, I see what you mean. I thought they were similarly styled but I see its pretty different. I agree then, we should change them all, I'm sure we can disable the time component of the new one. It's pretty configurable. I do think this should be done in a separate issue thought. It doesn't fit here. We could create an issue and make sure it gets done before the next release? How does that sound?

app/scripts/controllers/logs.js
rcrichton: Will fix.

rcrichton:
Cool, thanks guys.

@BMartinos I agree 0'ing the time would be better. I'l see if that can be done.

app/index.html
devcritter: Yeah, sounds good

app/scripts/controllers/logs.js
devcritter: I was thinking that if you leave the page open and let it keep auto appending it might make that string really big after a while - although to be fair I'm not sure if it'll actually be a problem in practice.

app/index.html
rcrichton: Ok here is the new issue: https://github.com/jembi/openhim-console/issues/338

app/scripts/controllers/logs.js
rcrichton: Right, I think that is what you are seeing @devcritter. I'm not really sure this is an issue as the clear button is intended to reset the filters. Would changing the button text from 'Clear' to 'Reset' be better?

app/scripts/controllers/logs.js
rcrichton: I could but I I didn't want to give the user too much choice. Also, autoupdate only functions when there isn't a date range set so would be disabled a lot of the time. Also with tailing off its easy to keep you place as logs update automatically as your view won't be reset. I don't mind changing it, but I'm happy with how it works atm. So, let me know how strongly you feel about this.

app/scripts/controllers/logs.js
devcritter: Aaah ok I see - cool yeah; or maybe 'Clear Filters'? (That's what it's currently called on the transactions page.)

rcrichton:
I've pushed some changes to address the comments.

* The time now 0's
* Changed 'Clear' to 'Reset' - is this better?
* I left the autoupdate thing, see comments above.
* I added some code to keep the logs string a reasonable size.
* Fixed a date format
* Added an issue for the changing the other datepickers to the new one.

Let me know what you think @devcritter?

app/scripts/controllers/logs.js
rcrichton: Clear Filters may be best of consistency. I'l update.

devcritter:
@rcrichton Cool looks good! :+1: The '0' time is really nice now. No objections if you want to leave the autoupdate.

PR #339, opened 2015-11-12
Transaction details page refactor

### LINKED ISSUE #303: Move primary route orchestrations out into their own section (not in response section)

### LINKED ISSUE #304: Move routes list out of response section and into its own section

### LINKED ISSUE #336: Enable easy scanning of responses
From @armageddon:

When investigating issues with completed transactions it is very laborious to check each message one by one.
It would be helpful to be able to click through the list of transactions selected by the filter and see immediately the relevant information.

This could be done as a pop up screen with backward and forward arrows , showing the response of the request .

rcrichton:
Closes #336 also closes #303 and closes #304 

I have decided to go with a solution that uses sessionStorage to stores an array of `_id`s for the current filtered transactions. The reason for this was two fold:

* It allows the transaction query to only happen once, not on every transaction details page load.
* It allowed me to not have to reimplement support for all of the filters that we currently have. The code there is not easily reused. I think that code need to be refactored a bit and I didn't want to proliferate it. (https://github.com/jembi/openhim-core-js/issues/589)

It seem to work pretty well so far, but I'm interested to know what other think.

@devcritter could you review this for me and @BMartinos could you have a look at some of the UI changes and let me know if there is anything I should do differently or if you have any better design ideas.

I also, cleaned up some of the css which was overriding some default bootstrap stuff (that isn't ideal). This may have some styling impact on other pages. I've had a look and tried to resolve these, but if you notice any others please let me know.

Also, @armageddon does this solve your problem for momconnect?

See some quick pics below:

image

image

image

### DIFF
--- a/app/views/transactionDetails.html
+++ b/app/views/transactionDetails.html
@@ -119,10 +119,13 @@
                 </div>
                 <div class="panel-body">
 
-                  <h4><span class="label label-default">{{transactionDetails.request.method}}</span>{{ ' ' + transactionDetails.request.host + ':' + transactionDetails.request.port + transactionDetails.request.path + '?' + transactionDetails.request.querystring}}</h4>
+                  <h4>
+                    <span class="label label-default" ng-if="channel.type === 'http'">{{transactionDetails.request.method}}</span>
+                    <span class="label label-danger" ng-if="channel.type !== 'http'">{{channel.type | uppercase}}</span>
+                    <span>{{ transactionDetails.request.host }}</span><span ng-if="transactionDetails.request.port">{{ ':' + transactionDetails.request.port }}</span><span ng-if="channel.type === 'http'">{{ transactionDetails.request.path }}</span><span ng-if="transactionDetails.request.querystring && channel.type === 'http'">{{ '?' + transactionDetails.request.querystring }}</span>
                   <h6 class="subtitle"><em>Received at {{transactionDetails.request.timestamp | date:'yyyy-MM-dd HH:mm:ss.sss Z'}}</em></h6>
 
-                  <a data-toggle="collapse" href="#req-headers-table">Show Headers <i class="glyphicon glyphicon-chevron-down"></i></a>
+                  <a data-toggle="collapse" href="#req-headers-table" ng-if="channel.type === 'http'">Show Headers <i class="glyphicon glyphicon-chevron-down"></i></a>
                   <div id="req-headers-table" class="panel panel-default collapse">
                     <div class="panel-body">
                       <table class="table-condensed table-borderless">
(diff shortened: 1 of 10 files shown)


app/views/transactionDetails.html
```diff
@@ -2,283 +2,266 @@
...
+                        <th>Route</th>
+                        <th>URL</th>
+                        <th>Status</th>
+                        <th>Response Time</th>
+                      </tr>
+                    </thead>
+                    <tbody>
+                      <tr class="table-list" ng-repeat="route in transactionDetails.routes" ng-click="viewAddReqResDetails(route)">
+                        <td data-title="Route">{{ route.name }}</td>
+                        <td data-title="URL">{{ route.request.method + ' ' + route.request.host + ':' + route.request.port + '/' + route.request.path + '?' + route.request.querystring }}</td>
```
devcritter: You should just add some existence checks for path/query - things like tcp routes won't have them.

devcritter: Maybe also indicate that it's tcp or mllp if appropriate

devcritter:
@rcrichton really nice! It works very well.

I don't have any issue with using the session storage, and it worked fine for me - the nice thing is it doesn't clutter the url and makes the url easier to share.

The route/orchestration popup is still using the old format - I believe it's [this](https://github.com/jembi/openhim-console/blob/master/app/views/transactionsAddReqResModal.html) one that needs to be updated as well [the request/response boxes].

I would also recommend just testing with a few tcp transactions (both primary and secondary) to see what it'll look like - I suspect there's a couple places that might look a bit odd.

I tried on a secondary quickly:
screen shot 2015-11-12 at 19 56 25

app/views/transactionDetails.html
rcrichton: Ok, good idea. I'l see what I can do.

rcrichton: Angular doesn't care if the values don't exist. It doesn't cause errors. Also, for TCP and MLLP rotues we don't have an indication that they are TCP from the transactions details. TCP transaction get converted to and from HTTP transactions internally. I'm not sure if they need to be displayed differently. What do you think?

The one problem that the current way has, is it will always append the ':' and '?' strings. I could try sort that out.

rcrichton: I've adjusted it to hide the method, path and query string on tcp as well as the headers. Does this seems ok:

image

devcritter: Yup that looks good! I think the conversion to http is an internal detail that we should rather abstract away from on the interface - but yeah this is good.

Yeah, that was what I meant about the existence checks: the dangling `:`s and `?`s.

rcrichton:
I've got TCP routes to display like this for secondary routes:

image

image

On the http side:

image

How does that look?

rcrichton:
@devcritter could you please review again?

devcritter:
Nice! yeah looks good :+1:

PR #340, opened 2015-11-12
Simple mapping

devcritter:
@rcrichton the companion PR for https://github.com/jembi/openhim-core-js/pull/590

screen shot 2015-11-12 at 20 24 57

I just reused you header box from the other PR, so it should look the same when that one is merged
screen shot 2015-11-12 at 20 24 26

### DIFF
--- a/app/views/mediatorConfigModal.html
+++ b/app/views/mediatorConfigModal.html
@@ -10,18 +10,19 @@
     <span>This mediator has does not have any config options that can be specified here.</span>
   </center>
 </div>
-<form name="form" novalidate>
+<form class="form-horizontal" ng-submit="saveConfig()" novalidate>
   <div class="modal-body">
     <fieldset>
 
-      <div class="form-group-2columns" ng-repeat="def in mediator.configDefs">
-        <div class="form-group" style="float: left; width: 33%;">
+      <div>
+        <div class="form-group" ng-repeat="def in mediator.configDefs">
+          <div class="col-sm-4">
           <label ng-if="def.displayName">{{def.displayName}}</label>
           <label ng-if="!def.displayName">{{def.param}}</label>
           <i ng-if="def.description" tooltip="{{def.description}}" class="glyphicon glyphicon-question-sign tooltip-wide"></i>
-        </div>
-        <div class="form-group" style="float: right; width: 67%">
+          </div>
 
+          <div class="col-sm-8">
           <div ng-if="def.type === 'string'">
             <input class="form-control" type="text" ng-model="mediator.config[def.param]">
           </div>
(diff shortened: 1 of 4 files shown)


app/views/mediatorConfigModal.html
```diff
@@ -15,19 +15,19 @@ <h4 class="modal-title" id="addMediatorConfigLabel">
     <fieldset>
 
       <div class="form-group-2columns" ng-repeat="def in mediator.configDefs">
-        <div class="form-group" style="float: left;">
+        <div class="form-group" style="float: left; width: 33%;">
```
rcrichton: It's better practise to not use inline styles. I know this was originally my code, but maybe we should switch to a horizontal-form instead of hard coding changed to the bootstrap styles - http://getbootstrap.com/css/#forms-horizontal

(I learnt a fair bit doing the transactions refactor, I didn't know about this stuff before...)

app/views/mediatorDetails.html
```diff
@@ -88,7 +88,22 @@
                       </strong>
                       <i ng-if="mediatorDefsMap[k].description" tooltip="{{mediatorDefsMap[k].description}}" class="glyphicon glyphicon-question-sign"></i>
                     </td>
-                    <td style="width: 75%">{{v}}</td>
+                    <td style="width: 75%">
```
rcrichton: Inline style.

app/views/mediatorConfigModal.html
```diff
@@ -38,11 +38,26 @@ <h4 class="modal-title" id="addMediatorConfigLabel">
...
+            <table style="width: 100%;">
+              <tr ng-repeat="(mapping, _v) in mediator.config[def.param]">
+                <td>{{mapping}}</td>
+                <td><input class="form-control" type="text" ng-model="mediator.config[def.param][mapping]"></td>
+                <td><button class="btn btn-danger btn-xs" ng-click="removeMapping(def.param, mapping)"><i class="glyphicon glyphicon-remove"></td>
+              </tr>
+              <tr>
+                <td><input class="form-control" type="text" ng-model="newMapping.key"></td>
+                <td><input class="form-control" type="text" ng-model="newMapping.value"></td>
+                <td><button class="btn btn-primary btn-xs" ng-click="addNewMapping(def.param)" ng-disabled="!newMapping.key"><i class="glyphicon glyphicon-plus"></td>
```
rcrichton: You may want to disable this if the key already exists in the map and provide some sort of indication of this.

rcrichton:
Nice, this looks good and will be pretty useful. A few comments above.

app/views/mediatorConfigModal.html
devcritter: Aah ok cool, will do.

devcritter:
@rcrichton made the changes as suggested - if you could please take another look?

app/views/mediatorDetails.html
```diff
@@ -81,14 +81,29 @@
...
                         <span ng-if="!mediatorDefsMap[k].displayName">{{k}}</span>
                       </strong>
                       <i ng-if="mediatorDefsMap[k].description" tooltip="{{mediatorDefsMap[k].description}}" class="glyphicon glyphicon-question-sign"></i>
                     </td>
-                    <td style="width: 75%">{{v}}</td>
+                    <td class="mediator-settings-col2">
+                      <div ng-if="mediatorDefsMap[k].type !== 'map'">{{v}}</div>
+                      <div ng-if="mediatorDefsMap[k].type === 'map'">
+                        <a data-toggle="collapse" href="#res-headers-table">Show Values <i class="glyphicon glyphicon-chevron-down"></i></a>
+                        <div id="res-headers-table" class="panel panel-default collapse">
```
rcrichton: You may want to change this id.

rcrichton:
@devcritter just a few more comments.

devcritter:
@rcrichton please take another look?

I had to use an id-bound style for the collapse-margins - the problem is that form-horizontal includes a `margin-left: -15` which is was trying to counteract in the first place, and this also overrides a class level style.

rcrichton:
Cool, looks good. You may be able to use a [`!important`](http://stackoverflow.com/a/9245360/588776) on the css class to get it to override correctly. I think that should work. Just, the toggle for the setting isn't working now with the id change, see comment above (I see you just fixed it).

After that :+1:

devcritter:
Aah cool, yeah the !important worked

PR #347, opened 2015-11-25
Config struct repeats

devcritter:
Companion PR for https://github.com/jembi/openhim-core-js/pull/600

@rcrichton if you could please review?

Structs:
screen shot 2015-11-25 at 13 17 32
screen shot 2015-11-25 at 13 17 20

Arrays:
screen shot 2015-11-25 at 16 05 33

### DIFF
--- a/app/views/partials/mediator-config.html
+++ b/app/views/partials/mediator-config.html
@@ -1,4 +1,5 @@
 <fieldset>
+  <div class="content-box-large bordered-box">
   <div class="form-group collapse-side-margins" ng-repeat="def in configDefs">
 
   <div ng-if="!def.array">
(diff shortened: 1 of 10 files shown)


devcritter:
screen shot 2015-11-25 at 16 21 52

rcrichton:
Hey @devcritter I got the following on the mediator details page when I add two empty structs and save.

```
Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: (arrKey, arrayValue) in configValue, Duplicate key: object:null
http://errors.angularjs.org/1.2.15/ngRepeat/dupes?p0=(arrKey%2C%20arrayValue)%20in%20configValue&p1=object%3Anull
    at angular.js:78
    at ngRepeatAction (angular.js:19729)
    at Object.$watchCollectionAction [as fn] (angular.js:11906)
    at Scope.$digest (angular.js:12031)
    at Scope.$apply (angular.js:12279)
    at done (angular.js:7991)
    at completeRequest (angular.js:8196)
    at XMLHttpRequest.xhr.onreadystatechange (angular.js:8137)
```

This was the config that I had:

image

This is what the mediator recieved back as config: `[{"weight":0.2,"host":"0.0.0.0"},{"weight":0.5,"host":"0.0.0.0"},{},{}]`

Perhaps there should be some protection against creating empty objects?

rcrichton:
Also, perhaps rather than displaying an empty box when struct has no values, we can display the text '_None_' or something like that.

app/views/partials/mediator-config.html
```diff
@@ -0,0 +1,141 @@
...
+    <div>
+      <div class="col-sm-3">
+        <div ng-if="!config[def.param] || config[def.param].length == 0">
+          <label ng-if="def.displayName">{{def.displayName}}</label>
+          <label ng-if="!def.displayName">{{def.param}}</label>
+          <i ng-if="def.description" tooltip="{{def.description}}" class="glyphicon glyphicon-question-sign tooltip-wide"></i>
+        </div>
+      </div><!-- col-sm-3 -->
+      <div class="col-sm-9 text-align-right">
+        <button type="button" class="btn btn-primary btn-xs" ng-click="addNewArrayItem(def)"><i class="glyphicon glyphicon-plus"/> <span ng-if="def.displayName">{{def.displayName}}</span><span ng-if="!def.displayName">{{def.param}}</span></button>
```
rcrichton: I think we need just a litle bit of top margin here to space this away from the box:

image

app/views/partials/mediator-config.html
```diff
@@ -0,0 +1,141 @@
...
+        <tr ng-if="doesNewKeyExist(def.param)">
+          <td><div class="has-error"><i class="glyphicon glyphicon-warning-sign"></i>Key already exists</div></td>
+          <td></td>
+          <td></td>
+        </tr>
+      </table>
+    </div>
+
+    <div ng-if="def.type === 'struct'">
+      <div mediator-nested-config config-defs="def.template" config="config[def.param]"></div>
```
rcrichton: This trick is great :)

rcrichton:
I feel like the config display for array structs need a bit more separation between the different struct objects.

image

Perhaps, each in their own box would be better or maybe even numbering each struct? Although this may look weird for non-struct arrays. Perhaps only if they are a struct they should be further grouped?

rcrichton:
@devcritter this looks great. I have added a few comment above.

devcritter:
@rcrichton I thought I fixed that issue already :( It's an issue with [duplicates](https://docs.angularjs.org/error/ngRepeat/dupes) in ng-repeat. Will investigate.

All good suggestions on the styling - will implement.

devcritter:
@rcrichton mind taking another look?

devcritter:
screen shot 2015-11-26 at 12 07 45

rcrichton:
:+1: - looks good to me. I found the directive code a bit hard to follow probably because I'm not that familiar with directives, but I think all looks good if I'm understanding correctly.

devcritter:
Cool, yeah, I'm new to directives as well so not 100% sure how good my code is. But will learn as we use these more.

rcrichton:
@bausmeier was telling me that apparently angular 2 make much more use of directives, so may be good to start trying to use them more in console.

bausmeier:
@rcrichton Angular 2 makes use of components which are similar to directives, but not the same. Not sure how similar they actually are.

PR #349, opened 2016-01-20
Config import: added null checks

devcritter:
@rcrichton quick one: just some checks in case some elements are left out of the import json


rcrichton:
Cool, looks good to me. :+1:

Issue #351, opened 2016-01-27
It should be possible to duplicate a channel


rcrichton:

rcrichton:
Closed with #360

PR #360, opened 2016-02-05
OHM-228: Channel Duplication functionality

BMartinos:
This PR relates to issue: #351

### DIFF
--- a/app/scripts/controllers/channelsmodal.js
+++ b/app/scripts/controllers/channelsmodal.js
@@ -303,19 +303,14 @@ app.controller('channelBasicInfoCtrl', function ($scope) {
     });
   };
 
-  // if update is true
-  if ($scope.update) {
+  // if update/channelDuplicate is true
+  if ($scope.update || $scope.channelDuplicate) {
     setUrlPattern();
   }else{
     // set default options if new channel
     $scope.channel.type = 'http';
     $scope.channel.authType = 'private';
     $scope.channel.status = 'enabled';
-
-    if ( $scope.channelDuplicate ){
-      setUrlPattern();
-    }
-
   }
 
 });
--- a/app/scripts/controllers/channelsmodal.js
+++ b/app/scripts/controllers/channelsmodal.js
@@ -2,7 +2,7 @@
 
 var app = angular.module('openhimConsoleApp');
 
-app.controller('ChannelsModalCtrl', function ($scope, $modalInstance, $timeout, Api, Notify, Alerting, channel) {
+app.controller('ChannelsModalCtrl', function ($scope, $modalInstance, $timeout, Api, Notify, Alerting, channel, channelDuplicate) {
 
   /****************************************************************/
   /**   These are the functions for the Channel initial load     **/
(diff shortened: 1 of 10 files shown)


BMartinos:
Hey @rcrichton Would you mind reviewing this PR for me when you get a chance. Thanks

app/scripts/controllers/channelsmodal.js
```diff
@@ -292,11 +301,21 @@ app.controller('channelBasicInfoCtrl', function ($scope) {
...
+  if ($scope.update) {
+    setUrlPattern();
   }else{
     // set default options if new channel
     $scope.channel.type = 'http';
     $scope.channel.authType = 'private';
     $scope.channel.status = 'enabled';
+
+    if ( $scope.channelDuplicate ){
+      setUrlPattern();
```
rcrichton: Does this need to be done here? Perhaps we could rather just make the if statement above `if ($scope.update || $scope.channelDuplicate) {`

app/views/channels.html
```diff
@@ -49,6 +49,7 @@
                   <td>
 
                     <div ng-if="channel.status!='deleted'">
+                      <button class="btn btn-info btn-xs" ng-click="duplicateChannel(channel)"><i class="glyphicon glyphicon-file"></i></button>
```
rcrichton: I think we need some tooltip text for this as it is not clear what it does from just looking at it.

While you are at it, perhaps adding tooltips to the two other buttons would be a good idea.

rcrichton:
@BMartinos thanks this works well! Just a few comments above.

Also, in terms of style I see you quite often write code like this:

`if ( $scope.channelDuplicate ){`

I think it's more common to write that like so

`if ($scope.channelDuplicate) {`

Just for consistency, you may want to write new code that that rather. But, not a major sticking point.

app/scripts/controllers/channelsmodal.js
BMartinos: Well spotted. Have updated in code.

app/views/channels.html
BMartinos: I originally added in the tooltips for the three button to explain the role of each. Visually, this didn't look so great to me as there were too many tooltip popups for the buttons as i hovered over them. I can add them in to illustrate? or an alternative is to added the description in the column heading E.g "Duplicate/Edit/Delete"

BMartinos:
Hey @rcrichton Thanks for the feedback. I've normally just adopted the extra space style for my code as I find it easier to read and track but will remove spaces :). Just one question regarding tooltips before i commit changed code. Thanks

app/views/channels.html
rcrichton: Ok I see. I think tooltips would be ok. Even if they just say 'duplicate', 'edit' and 'delete'. I'm not such a fan of adding these to the header, it's not very clear. If you could provide a screenshot of what you think looks bad we could try iterate on that.

BMartinos:
Hey @rcrichton Ive added in the tooltips for those action buttons. Its a bit hard to explain the UI issue with a screenshot as you need to hover over a few of them to see it, but this might just look weird to me though. Have a look and let me know what you think. Thanks

rcrichton:
Hey @BMartinos, sorry I forgot toe reply to the PR.

I see there is a delay attribute on a tooltip. That may be useful. Perhaps we could display the tooltip only after 500ms: http://getbootstrap.com/javascript/#tooltips

Also, I think it would look better if there if the width of the tooltip matched the text.

rcrichton:
I think this is the issue with the tooltip style: https://github.com/jembi/openhim-console/blob/master/app/styles/main.css#L1256

It would be better practise to have a more specific selector or a separate class to change the width rather than changing it globally for all tooltips.

rcrichton:
@BMartinos I took this on due to your training. I think it is better now, what do you think? If you agree then I think this is a :+1:

BMartinos:
Hey @rcrichton thanks for picking this up, I think it looks much better now with the smaller width and the delay also makes the interaction a lot easier and user friendly. This is good to go :+1:

Issue #352, opened 2016-01-27
Channel Order


hnnesv:
See https://github.com/jembi/openhim-core-js/issues/612

Channels in the channel list should be ordered. In addition, add functionality that easily allows channels to be re-ordered (e.g. click and move)

PR #363, opened 2016-02-08
OHM-230: Added functionality to set channel priority level

BMartinos:
This PR relates to issue: #352 

The terminology has been changed from "Order" to "Priority" as this is more descriptive in terms of its purpose.

This PR requires an additional property in the channels' schema ( https://github.com/jembi/openhim-core-js/blob/master/src/model/channels.coffee#L33 )
"priority":           type: Number, min: 1

### DIFF
--- a/app/scripts/controllers/channels.js
+++ b/app/scripts/controllers/channels.js
@@ -3,7 +3,6 @@
 angular.module('openhimConsoleApp')
   .controller('ChannelsCtrl', function ($scope, $modal, Api, Alerting) {
 
-
     /* -------------------------Initial load & onChanged---------------------------- */
     var querySuccess = function(channels){
       $scope.channels = channels;
@@ -136,6 +135,44 @@ angular.module('openhimConsoleApp')
     var restoreError = function (err) {
       // add the error message
       Alerting.AlertAddMsg('top', 'danger', 'An error has occurred while restoring the channel: #' + err.status + ' - ' + err.data);
+    };
+
+
+
+    /*--------------------------Update priority Level----------------------------*/
+    $scope.updateChannelPriority = function(channel, direction){
+      var newPriority;
+      var curPriority = channel.priority;
+      if ( !curPriority ){
+        newPriority = getLowestPriority() + 1;
+      }else{
+        // set priority to lower number ( minus 1 )
+        if ( direction === 'up' ){
+          if ( curPriority === 1 ){ return; }
+
+          newPriority = curPriority - 1;
+        }else{
+          newPriority = curPriority + 1;
+        }  
+      }
+
+      Api.Channels.update({ _id: channel._id, priority: newPriority }, function(){
+        // reload channels
+        $scope.$broadcast('channelsChanged');
+      }, function( err ){
+        console.log( err );
+      });
+
+    };
+
+    var getLowestPriority = function(){
+      var lowestPriority = 0;
+      for (var i = 0; i<$scope.channels.length; i++){
+        if ( $scope.channels[i].priority && $scope.channels[i].priority > lowestPriority){
+          lowestPriority = $scope.channels[i].priority;
+        }
+      }
+      return lowestPriority;
     };
 
   });
--- a/app/scripts/controllers/channels.js
+++ b/app/scripts/controllers/channels.js
@@ -144,7 +144,7 @@ angular.module('openhimConsoleApp')
       var newPriority;
       var curPriority = channel.priority;
       if ( !curPriority ){
-        newPriority = getLowestPriority() + 1;
+        newPriority = $scope.getLowestPriority() + 1;
       }else{
         // set priority to lower number ( minus 1 )
         if ( direction === 'up' ){
@@ -160,12 +160,12 @@ angular.module('openhimConsoleApp')
         // reload channels
         $scope.$broadcast('channelsChanged');
       }, function( err ){
-        console.log( err );
+        Alerting.AlertAddMsg('top', 'danger', 'An error has occurred while updating the channel: #' + err.status + ' - ' + err.data);
       });
 
     };
 
-    var getLowestPriority = function(){
+    $scope.getLowestPriority = function(){
       var lowestPriority = 0;
       for (var i = 0; i<$scope.channels.length; i++){
         if ( $scope.channels[i].priority && $scope.channels[i].priority > lowestPriority){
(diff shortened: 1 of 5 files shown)


BMartinos:
Hey @hnnesv Would you mind reviewing this PR for me when you get a chance. Thanks

app/views/channels.html
```diff
@@ -14,32 +14,55 @@
...
+            <table class="table table-striped channelListingTable">
               <thead>
                   <tr>
                     <th width="50">#</th>
                     <th>Channel Name</th>
-                    <th>Channel Description</th>
                     <th>Channel URL</th>
+                    <th>
+                      Priority
+                      <i ng-show="uiSettings.showTooltips" tooltip-html-unsafe="Lower Value = Higher Priority<br />Higher Value = Lower Priority" class="glyphicon glyphicon-question-sign" style="font-size: 12px"></i>
```
hnnesv: Let's replace the tooltip with the following text:
> If a transaction gets matched to two or more channels, then it will get routed to the channel with the highest priority. A value of 1 is the highest possible priority and larger numbers indicate a lower priority.

app/views/partials/channels-tab-basic-info.html
```diff
@@ -6,22 +6,30 @@
...
+    <div class="has-error-msg" style="clear: both; margin-top: 5px;"><i class="glyphicon glyphicon-warning-sign"></i>{{validationRequiredMsg}}</div>
+
   </div>
-  <div class="addUrlRegexDelimiterDiv">
-    <label>
-      <input type="checkbox" ng-model="urlPattern.regex" style="margin-top: 0px; vertical-align: middle;" /> Auto-add<br />regex delimiters
-      <i ng-show="uiSettings.showTooltips" tooltip="This automatically adds start and ending anchors to the URL pattern regex. This is generally what you want." class="glyphicon glyphicon-question-sign" style="font-size: 12px"></i>
-    </label>
+  <div class="form-group" style="float: right; width: 23%;">
+    <label>Channel Priority</label>
```
hnnesv: Can you please add the tooltip help here as well?

hnnesv:
@BMartinos thanks, it works well!

Just some minor comments. Also please keep this PR open until the core changes are merged in (https://github.com/jembi/openhim-core-js/pull/617)

rcrichton:
One issue that I noticed is that disabled channels aren't at the bottom anymore. I think these should still be at the at the end of the list no matte what their priority is.

rcrichton:
Hey @BMartinos @hnnesv the initial functionality looks good. However, I think it's a bit awkward to the way the arrows work. With it being next to the priority number it seems strange sometime when you click 'up' and the number goes down and row moves up (it makes sense put is awkward). I think we need to introduce a little disconnect between the number and the up and down arrows.

One way that I'm thinking of is we move the up and down arrow to the start of the row outside of the table so that they seem connected to the row. Then perhaps we should replace the current '#' column (which serves no purpose) with the priority. Also, instead of having the up and down arrow on top of each other perhaps have them next to each other. The on-top-of-each-other style is often used for increasing and decreasing numbers, we would want to avoid that. Perhaps using these icons instead:
up down

What do you guys think?

hnnesv:
Yeah that sounds good!

BMartinos:
ok great ill make these changes. I agree, the numbering up/down with regards to the level is a lil confusing at first cause up is down and visa versa :). I think its a good idea to disconnect them and will cause less confusion to the user.

BMartinos:
Hey @rcrichton Ive made the changes mentioned above if you can have a look and let me know what you think. Thanks

app/scripts/controllers/channels.js
```diff
@@ -138,4 +137,42 @@ angular.module('openhimConsoleApp')
...
+        }else{
+          newPriority = curPriority + 1;
+        }  
+      }
+
+      Api.Channels.update({ _id: channel._id, priority: newPriority }, function(){
+        // reload channels
+        $scope.$broadcast('channelsChanged');
+      }, function( err ){
+        console.log( err );
```
rcrichton: Should we not also add an alert to the user here?

app/scripts/controllers/channels.js
```diff
@@ -138,4 +137,42 @@ angular.module('openhimConsoleApp')
...
+      });
+
+    };
+
+    var getLowestPriority = function(){
```
rcrichton: Please add a unit test for this.

app/scripts/controllers/channels.js
```diff
@@ -138,4 +137,42 @@ angular.module('openhimConsoleApp')
       Alerting.AlertAddMsg('top', 'danger', 'An error has occurred while restoring the channel: #' + err.status + ' - ' + err.data);
     };
 
+
+
+    /*--------------------------Update priority Level----------------------------*/
+    $scope.updateChannelPriority = function(channel, direction){
```
rcrichton: Please add a unit test for this.

rcrichton:
@BMartinos this new style looks much better! Nice. We just need some unit tests and I have a small comment above.

BMartinos:
Hey @rcrichton I have made the corrections as mentioned above. Can you review it again for me please. Thanks

test/spec/controllers/channels.js
```diff
@@ -138,4 +138,47 @@ describe('Controller: ChannelsCtrl', function () {
...
+
+    scope.updateChannelPriority(scope.channels[2], 'down');
+    scope.updateChannelPriority(scope.channels[2], 'down');
+    scope.updateChannelPriority(scope.channels[2], 'down');
+    scope.channels[2].should.have.property('priority', 10);   
+
+    scope.channels[0].$update.should.be.called;
+    scope.channels[1].$update.should.be.called;
+    scope.channels[2].$update.should.be.called;
+  });
```
rcrichton: Side note: usually unit tests should test only one thing at a time, so this should be broken into multiple tests for each thing it's testing. You could abstract out the common setup stuff into a function and use that in the other tests to make it easier to repeat the tests.

I don't think you need to change this here, it works well as it is as they are pretty straight forward. Just an FYI. :)

rcrichton:
Awesome, looks good :+1:

BMartinos:
Great thanks @rcrichton. Will make a note of it for future tests :). @hnnesv When you are ready to merge core ill merge in console

hnnesv:
on 3... :)

PR #365, opened 2016-02-09
OHM-225 Issue #364: Routes enable/disable

### LINKED ISSUE #364: Disable/enable routes
Console counterpart to https://github.com/jembi/openhim-core-js/issues/613

hnnesv:
Closes #364 

@rcrichton @BMartinos if you could both please take a look? thanks

screen shot 2016-02-09 at 11 37 52
screen shot 2016-02-09 at 11 38 04

### DIFF
--- a/app/views/partials/channels-tab-routes.html
+++ b/app/views/partials/channels-tab-routes.html
@@ -23,7 +23,7 @@
         <label>Status</label><br />
         <div class="btn-group" style="clear: both;">
           <label class="btn btn-primary" ng-model="newRoute.status" btn-radio="'enabled'">Enabled</label>
-          <label class="btn btn-primary" ng-model="newRoute.status" btn-radio="'disabled'">Disabled</label>
+          <label ng-click="onRouteDisable(newRoute)" class="btn btn-primary" ng-model="newRoute.status" btn-radio="'disabled'">Disabled</label>
         </div>
       </div>
     </div>
--- a/app/views/partials/channels-tab-routes.html
+++ b/app/views/partials/channels-tab-routes.html
@@ -12,10 +12,19 @@
       <div class="has-error-msg"><i class="glyphicon glyphicon-warning-sign"></i>{{validationRequiredMsg}}</div>
     </div>
     <div class="form-group" style="float: left; margin-left: 20px;">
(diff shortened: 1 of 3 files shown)


BMartinos:
Hey @hnnesv i see you just added the property "status: 'enabled'" when adding a new route, I think this needs to be done when adding a mediator route as well ( https://github.com/jembi/openhim-console/blob/49f76706a7885f20c178e0494b1a3c7703322899/app/scripts/controllers/channelsmodal.js#L528 ). Other than that this looks good to me :+1:

hnnesv:
Thanks, good catch

rcrichton:
I'm just going through the code now, but testing the UI, I think that when a route is marked as disabled it should be removed as a primary route. Else you may end up with multiple primaries if you re-enable a primary. Our error handling stops this from being saved, but removing something as a primary when disabled kind makes sense to me.

BMartinos:
:+1: on @rcrichton comment above. When a route is disabled it should remove the primary flag from that route.

app/views/partials/channels-tab-routes.html
```diff
@@ -12,10 +12,19 @@
...
-        <label class="btn btn-primary" ng-model="newRoute.primary" btn-radio="true">True</label>
-        <label class="btn btn-primary" ng-model="newRoute.primary" btn-radio="false">False</label>
+      <div style="float: left;">
+        <label>Primary Route</label><br />
+        <div class="btn-group" style="clear: both;">
+          <label class="btn btn-primary" ng-model="newRoute.primary" btn-radio="true">True</label>
+          <label class="btn btn-primary" ng-model="newRoute.primary" btn-radio="false">False</label>
+        </div>
+      </div>
+      <div style="float: left; margin-left: 20px;">
```
rcrichton: I'm not a fan of inline styles, but I see they are everywhere in this page. The float you could replace with a pull-left bootstrap calls. I'm not sure the best way to do the spacing though.

[non-blocking PR comment]

rcrichton:
@hnnesv Looks good, some comments above.

hnnesv:
@rcrichton added in the change if you could take another look?

rcrichton:
Looks good to me :+1:

PR #367, opened 2016-02-15
Issue #357: Informative error message on connection failure

### LINKED ISSUE #357: User friendly self-signed certificate warning message
The most common error encountered on the console is when login fails due to a self-signed cert. However the error message is too generic and confusing for new users. If possible, the console should detect this error and provide an appropriate message. Perhaps instructions could also be displayed helping the user resolve the problem (with a link to the api where they need to accept the cert from)

hnnesv:
Closes #357 

@rcrichton @BMartinos I thought it'd be great to get this in before the next release.

Unfortunately I can't see a good way to specifically determine that it's a cert issue (and seems to be a general [problem](http://stackoverflow.com/questions/31058764/determine-if-ajax-call-failed-due-to-insecure-response-or-connection-refused)) so just wrote it up generically

### DIFF
--- a/app/views/login.html
+++ b/app/views/login.html
@@ -46,9 +46,9 @@
       </div>
 
       <div ng-if="coreConnectionError" class="alert alert-danger" style="padding: 10px; margin: 15px 10%;">
-          An error occurred while attempting to connect to the OpenHIM Core.<br><br>
+          An error occurred while attempting to connect to the OpenHIM Core on https://{{config.host}}:{{config.port}}.<br><br>
           The service may not be running or may not be accessible from you current location.<br><br>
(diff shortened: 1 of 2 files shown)


hnnesv:
screen shot 2016-02-15 at 10 37 25

BMartinos:
:+1: looks good to me

app/views/login.html
```diff
@@ -45,8 +45,15 @@
         </div>
       </div>
 
+      <div ng-if="coreConnectionError" class="alert alert-danger" style="padding: 10px; margin: 15px 10%;">
+          An error occurred while attempting to connect to the OpenHIM Core.<br><br>
```
rcrichton: Could we perhaps list the connection URL that the API attempted to connect to?

rcrichton:
Looks good, just a small comment above.

hnnesv:
@rcrichton ok cool: 
screen shot 2016-02-15 at 12 45 08

rcrichton:
Cool, looks good to me :+1:

PR #368, opened 2016-03-02
Prevent future audit dates repeating

### LINKED ISSUE #350: Audit log - repeating audit when future event date is used
When an audit is received with an event time that is in the future, the auto-update mechanism keep repeating that audit every time it updates.

repeat-audit

napiergit:
@hnnesv can you please review.  Fixes #350


hnnesv:
:+1: Looks good to me!

PR #374, opened 2016-03-18
Add roles management

napiergit:
Hi @hnnesv, I wrote this using only the getRoles endpoint.  You can check it out to see if you like how it looks and feels.  There are still a number of changes I would like to make:

Roles
  Presently the api only returns roles that are assigned to channels
      Add functionality to return roles assigned to clients (@hnnesv) ?
      If role is removed from all channels it remains in client object
  Don’t display new role name on add role click if it already exists i.e. Role0
  Get Alerting Correct
      Alert for delete role (not for renaming role)
  Edit role name
      Slow update time when saved
  Tests
  Error handling
  Comments
  Refactor to use improved api-endpoints


hnnesv:
Thanks @napiergit it's looking good so far - I like the look and feel of it.

The roles API is finished now and I think it would address several of your listed issues.

One note so far; instead of making this a separate page, I think we should integrate it into the clients page, the same way the users channel management is a subitem in the users page.

rcrichton:
Hey @napiergit thanks this is looking good. Just a few comments. I can't seems to create a new role at the moment using the '+ Role' button. I get the following in the logs:

`2016-04-04T13:25:34.114Z - error: [worker1] Could not add a role via the API: Cannot read property 'length' of undefined`

and a weird json print out at the bottom of the clients page.

clients

Also, my natural reaction when editing a role name was to press enter when done, perhaps we should allow this.

Finally, it wasn't obvious to me that the client in a role were toggle-able at first, perhaps including a checkbox in front of each client when editing them could show the user that they are clickable.

Do you have some space to make these changes?

rcrichton:
I see that enter does actually work, however I had some toruble with it at first... I quite sure why.

rcrichton:
One more minor comment. We could use some tooltips on the two edit icons that explain what can be edited if you click them.

napiergit:
Hi @rcrichton, thanks for the feedback.  I hsould have time later in this sprint to make the changes.

app/index.html
```diff
@@ -69,8 +69,8 @@
                             <li ng-class="{current: isCurrent('/')}"><a href="#"><i class="glyphicon glyphicon-home"></i> Dashboard</a></li>
                             <li ng-class="{current: isCurrent('/transactions')}"><a href="#/transactions"><i class="glyphicon glyphicon-list-alt"></i> Transaction log</a></li>
                             <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/audits')}"><a href="#/audits"><i class="glyphicon glyphicon-book"></i> Audit Log</a></li>
-                            <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/clients')}"><a href="#/clients"><i class="glyphicon glyphicon-th"></i> Clients</a></li>
                             <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/channels')}"><a href="#/channels"><i class="glyphicon glyphicon-random"></i> Channels</a></li>
+                            <li ng-show="userGroupAdmin" ng-class="{current: isCurrent('/clients')}"><a href="#/clients"><i class="glyphicon glyphicon-th"></i> Clients</a></li>
```
rcrichton: I'm not sure why the order has been changed, perhaps this is a refactor mistake?

app/scripts/controllers/clients.js
```diff
@@ -55,21 +55,327 @@ angular.module('openhimConsoleApp')
...
+          buildClientsRolesObject();
+          buildChannelsRolesObject();
+          angular.forEach($scope.roles, function(role) {
+            role.displayName = role.name;
+          });
+        }
+      });
+    };
+    
+    var buildClientsRolesObject = function() {
```
rcrichton: It's better practise to define a function above where it is used in code. Here, it works fine because the function is only called when $watch fires, however, perhaps we should still move the code for stylistic reasons? (This is done above as well).

app/styles/main.css
```diff
@@ -1369,3 +1369,21 @@ label { position: relative; }
...
+
+.rolesChannelsMatrixDiv .isAllowed { color: #009933; margin: 0px 2px; font-size: 13px; }
+.rolesChannelsMatrixDiv .isNotAllowed { color: #FF0000; margin: 0px 2px; font-size: 13px; }
+
+.limitChannelTdWidth {width:120px;}
+
+.roleAssignedToClientSpan { color: #333333; }
+.roleNotAssignedToClientSpan { color: #999; text-decoration: line-through; }
+
+.cursorPointer {cursor: pointer}
```
rcrichton: Just a style comment, please put spaces after your `{` as done above so it stays consistent.

app/views/clientsmodal.html
```diff
@@ -62,7 +62,12 @@ <h4 class="modal-title" id="addClientLabel">
 
       <div class="form-group" ng-class="{ 'has-error' : ngError.roles }">
         <label><span>*</span> Roles</label>
-        <taglist tag-data="client.roles"><input class="form-control" placeholder="Enter a role and press enter..." type="text" name="roles" ng-model="roles" typeahead="role for role in taglistClientRoleOptions | filter:$viewValue" /></taglist>
+        <br/>
+        <span ng-repeat="role in selectClientRoleOptions">
+          <input type="checkbox" ng-model="bleh" ng-checked="assigned[role]" ng-change="toggleRoleAssignedToClient(role)">{{role}}</input>
```
rcrichton: 'bleh' :)

What does this do exactly?

app/views/sidebar.html
```diff
@@ -8,8 +8,8 @@
       <li ng-class="{current: isCurrent('/')}"><a href="#"><i class="glyphicon glyphicon-home"></i> Dashboard</a></li>
       <li ng-class="{current: isCurrent('/transactions')}"><a href="#/transactions"><i class="glyphicon glyphicon-list-alt"></i> Transaction log</a></li>
       <li ng-if="userGroupAdmin" ng-class="{current: isCurrent('/audits')}"><a href="#/audits"><i class="glyphicon glyphicon-book"></i> Audit log</a></li>
-      <li ng-if="userGroupAdmin" ng-class="{current: isCurrent('/clients')}"><a href="#/clients"><i class="glyphicon glyphicon-th"></i> Clients</a></li>
       <li ng-if="userGroupAdmin" ng-class="{current: isCurrent('/channels')}"><a href="#/channels"><i class="glyphicon glyphicon-random"></i> Channels</a></li>
+      <li ng-if="userGroupAdmin" ng-class="{current: isCurrent('/clients')}"><a href="#/clients"><i class="glyphicon glyphicon-th"></i> Clients</a></li>
```
rcrichton: Again, does this order need to be changed?

rcrichton:
Hey @napiergit, I've done a first pass at code review. It's looking good! Nice work. I'll still need to do another pass once this is updated as I didn't grasp everything. Also you may want to look into writing some test for the functions that you wrote.

app/views/clientsmodal.html
napiergit: Hi @rcrichton . I want the functionality of the ng-change directive, which requires an ng-model for some arbitrary reason, so I created a dummy.  Couldn't think of a better name offhand :)

rcrichton: Ok cool.

napiergit:
Hi @rcrichton , can you please review if you have a chance.

rcrichton:
Hi @napiergit, thanks, I've played around with the console and it works well. Just a few comments from before that I think you missed:

1. Perhaps add a checkbox in front of the clients to show they can be toggled, else it's unclear to the user at first.
2. Put some tooltips on the yellow edit icons explaining what they do. (eg. 'Edit the clients in each role. If all clients and channels are removed from a role it will be deleted.')

Otherwise, it works great. I'l dig into the code now to review properly but you can work on these minor things in the meantime.

rcrichton:
@napiergit also, you may want to start merging master into this branch as currently there are conflicts.

napiergit:
Hi @rcrichton, the tooltips should already be there, but I'll change them to the above.  With regard to the checkbox, is clicking on the "Edit clients in each role" and thereafter the cursorPointer hover not enough to make a user feel they are editable?

rcrichton:
Ok cool, ya I just saw the tooltips. Those look good. I think we can leave the checkboxes, it should be easy enough to understand now.

app/scripts/controllers/clientsmodal.js
```diff
@@ -8,23 +8,90 @@ angular.module('openhimConsoleApp')
...
+          $scope.formData.duplicateNewRole = true;
+        } else {
+          $scope.formData.duplicateNewRole = false;
+          $scope.client.roles.push(newRole);
+          $scope.roles.push({name: newRole});
+          $scope.formData.assigned[newRole] = true;
+          $scope.formData.newClientRole = null;
+        }
+      }
+    };
```
rcrichton: A lot of this code looks very similar to the code in `channelsmodal.js` above. If it is repeated (it's hard to tell just by looking at it) then it would be much better to split this functions out into another file that is shared between these controllers. Otherwise, this is breaking the DRY (Don't repeat yourself) programming principle. Would it be possible to split it out?

napiergit: It certainly would be possible, my only hesitation is: is the code relevant enough to be given its own service?

app/views/partials/channels-tab-access-control.html
```diff
@@ -1,7 +1,28 @@
...
+      <br/>
+    </span>
+    <div class="has-error-msg"><i class="glyphicon glyphicon-warning-sign"></i>Either '<strong>Assign Existing Roles</strong>' or '<strong>Add New Role</strong>' is required!</div>
+  </div>
+  <div class="form-group" style="float: right;" ng-class="{ 'has-error' : ngError.roles }">
+    <label><span>**</span> Add New Role</label>
+    <input class="form-control" type="text" name="newRole" ng-model="formData.newChannelRole" ng-blur="createNewRole()" ng-keyup="$event.keyCode == 13 && createNewRole()"/>
+    <div style="color:orange" ng-show="formData.duplicateNewRole"><i class="glyphicon glyphicon-warning-sign"></i>Role/Client already assigned to Channel</div>
+    <div class="has-error-msg"><i class="glyphicon glyphicon-warning-sign"></i>Either '<strong>Assign Existing Roles</strong>' or '<strong>Add New Role</strong>' is required!</div>
+  </div>
```
rcrichton: I'm not able to directly add a clients anymore. I think we need this functionality back.

test/spec/controllers/clients.js
```diff
@@ -73,4 +87,158 @@ describe('Controller: ClientsCtrl', function () {
...
+    scope.newRoles[0] = {name:'test'};
+    scope.newRoles[1] = {name:'test2'};
+    
+    scope.saveNewRole(role);
+    httpBackend.flush();
+    scope.newRoles.length.should.equal(1);
+    scope.newRoles[0].should.have.property('name', 'test2');
+  });
+  
+  it('should remove a new roll', function () {
```
rcrichton: typo: roll -> role

app/views/partials/channels-tab-access-control.html
napiergit: We want to move away from adding clients directly to channels.  You can still add a client by typing its name in the add new role input.

rcrichton:
@napiergit thanks this is looking good, I've added a few comments above that you should address. Also, I think that you should focus on merging in master to this as the channels modal has changed a fair bit since this branch was first created. Let me know if you need any help or have questions about this. Also, the tests are looking nice!

napiergit:
Hi @rcrichton, I have given my thoughts on some of the issues you raised.  If there is anything that you would like me to do regardless of my feedback, please let me know and I'll jump on it.  I will be fine with the merge conflicts thanks.

app/scripts/controllers/clientsmodal.js
rcrichton: I think it is. Just for maintainability reasons (you don't have to change the code in two places). Another way to add those to the utils script - https://github.com/jembi/openhim-console/blob/master/app/scripts/util/utils.js

rcrichton:
@napiergit cool, see my responses above.

rcrichton:
Oh, also please rename this PR to remove the 'Do no merge' text :) - we will be merging this soon.

app/views/partials/channels-tab-access-control.html
rcrichton: I still this it's good to allow this for simple use cases. Typing in a name directly would be very error prone. It should be a larger discussion if we decide that we don't want to support that.

napiergit:
Hi @rcrichton, I fixed our DRY problem and adding clients on channels problem by removing my footprint from the channelsmodal controller.  That should be about it.

app/scripts/controllers/channelsmodal.js
```diff
@@ -1199,3 +1198,15 @@ app.controller('channelAlertsCtrl', function ($scope, Api) {
   /****************************************************************/
 
 });
+
+// nested controller for the channel settings tab
+app.controller('channelSettingsCtrl', function ($scope) {
```
rcrichton: I think this controller need to be removed, it seems to be a repeat from line 783. I may have been a mistake in the reverting.

rcrichton:
@napiergit cool, just _one_ more comment. I think that code needs to be removed, otherwise this looks good. It's got conflicts again because of the other PR of your that I just merged :( but those should be easy to fix.

napiergit:
Cool @rcrichton, I've sorted the merge conflicts again.  Just waiting for travis.

rcrichton:
Cool, looks good :+1: Once travis is done I think we can merge.

napiergit:
Yay!!

PR #376, opened 2016-03-23
Issue #375: Always show users-channels matrix

### LINKED ISSUE #375: Don't hide users to channels subcomponent
Currently on the Users page, the users to channels permissions subcomponent is hidden by default and only shown by clicking on a show/hide button. This button should be removed and the component always shown.
screen shot 2016-03-23 at 09 54 10

hnnesv:
Closes #375 

@BMartinos does this look okay?


hnnesv:
@BMartinos ping

You wouldn't mind just looking at this quickly?

BMartinos:
Hey @hnnesv sorry this slipped by me. Looks good to me :+1: 
Wondering if it might not be a good idea to link the user email in the legend list as well?

hnnesv:
Cool yeah, good idea, added it in. At some point we could even consider refactoring the two boxes into a single one - i.e. a single box with the permissions and user details.

hnnesv:
@BMartinos actually it was a easy enough change. Mind taking another glance?
screen shot 2016-03-30 at 09 52 35

BMartinos:
Nice one! looks good to me. Just a small one, maybe we can combine the firstname/lastname into a single column called "Fullname", just to save some space for when there are a lot of channels. But otherwise :+1:

PR #379, opened 2016-03-30
Issue #378: Fixed relative paths

### LINKED ISSUE #378: Transaction details doesn't take non-root basepath into account
The prev/next buttons on the transaction details page doesn't take it into account if the console is hosted on a path other than /
https://github.com/jembi/openhim-console/blob/master/app/views/transactionDetails.html#L76

hnnesv:
Closes #378 

@BMartinos mind taking a look at this with Ryan being away? thanks


BMartinos:
Hey @hnnesv this looks good to me. Think you got all of them, couldn't find any more doing a global search :+1:

PR #386, opened 2016-05-30
Issue #385: Set data model to use channel for route type checks

### LINKED ISSUE #385: Transaction transactionsAddReqResModal not showing headers correctly
The transactions req/res modal box isn't displaying the headers:
screen shot 2016-05-30 at 13 08 09

The code is present [though](https://github.com/jembi/openhim-console/blob/master/app/views/transactionsAddReqResModal.html#L31-L41) and the headers can be seen in the html itself on the page. The show headers link might just not be visible...

hnnesv:
Closes #385 

@rcrichton if you could please review? Refactored the code; not quite sure about what was there originally but it didn't really make sense here, which is why it wasn't working correctly - must've been a casualty of refactoring or something.

### DIFF
--- a/app/scripts/controllers/transactionDetails.js
+++ b/app/scripts/controllers/transactionDetails.js
@@ -210,7 +210,7 @@ angular.module('openhimConsoleApp')
     /**               Transactions View Route Functions                 **/
     /*********************************************************************/
 
-    $scope.viewAddReqResDetails = function(record, route){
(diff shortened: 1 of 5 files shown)


rcrichton:
Ah ok awesome. Seems to work well for we. (Took me a while to find where this UI actually was though :) ) :+1:

PR #388, opened 2016-06-22
DAT-22 Add option to set Authorization header forwarding in routes

rcrichton:


hnnesv:
Tested and it works well! I just found it a bit confusing having this and the username/password fields on the route panel, i.e. it isn't really clear from a UI pov what happens if you set it to true and have other basic auth details for the route (i saw of course when running a transaction that the auth forwarding takes precedence over the route's basic auth details).

How about we just disable the username/password fields if forward is set to true? that way it's immediately clear to the user that auth forwarding will take place, regardless of what value is set for username/password?

hnnesv:
Another super-minor point - maybe set forward auth to false for a new channel? rather then unselected?

rcrichton:
@hnnesv ok cool, I'l work on those.

rcrichton:
@hnnesv please review again.

hnnesv:
👍

PR #397, opened 2016-07-26
Ohm 266 fix email bugs

### LINKED ISSUE #348: Config Import shouldn't email users
Currently when importing a config export from another HIM instance, imported users will be locked and they will be emailed with a password reset even though their password hashes were set correctly.

(this may need core changes rather than console)

napiergit:
fixes #348


test/spec/controllers/exportimport.js
```diff
@@ -210,7 +203,7 @@ describe('Controller: ExportImportCtrl', function () {
 
     scope.createExportFile( scope.selectedExports );
 
-    expect(scope.downloadLink.indexOf('blob:http://localhost:8080/')).to.be.above(-1);
+    expect(scope.downloadLink.indexOf('')).to.be.above(-1);
```
napiergit: When running the tests, the createExportFile method fails with the error "browser not capable of creating blob", hence no download link.  So will have to reconsider this.  Thoughts?

hnnesv: This is something that came up when we upgraded the packages last week. Older versions of phantom didn't support blobs, the newer version does. Please try deleting your node_modules folder, npm install and then running the tests again?

app/scripts/controllers/exportImport.js
```diff
@@ -199,111 +202,44 @@ angular.module('openhimConsoleApp')
...
-          $scope.importProgressStatus = Math.floor( doneItems / totalRecords );
-
-          // update progress bar too 100%
-          if( doneItems === totalRecords ){
-            $scope.importProgressStatus = 100;
-            $scope.importProgressType = 'success';
-          }
-
+      
+      // If record has _id then do update instead of insert
```
nthfloor2: Looking good so far! Just a quick question though regarding the 'update script' - what is the intended use case for this? Since we are exporting the Mongo _ids which are only relevant to this DB, how would someone use this feature?

At the moment I can think of of perhaps making backups of the current config, just interested to know if there are other reasons...

napiergit: Good question.  That's the only reason I can think of at the moment too.

nthfloor2:
Awesome, looks good!

One thing I found is that when I import an update script after changing the password for one of the user accounts, the password is not reset. Do you want to be the case or should the update script reset the password to what is was when the script was generated?

app/scripts/controllers/exportImport.js
nthfloor2: Ok cool, let's leave as is.  I think it could come in handy especially when we are playing around with different mediators and configs. Probably not that useful to a user, since the config will not likely change much once they have setup OpenHim on their system. 

@rcrichton Thoughts?

rcrichton: Persisting the `_id` allows us to update certain documents rather than just adding new ones. This is helpful when importing config into a database without duplicating things that exist. But yes it will only work for the same database where the `_id`'s match. Backups are a good example use case. It's not the cleanest solution but I'm not sure of another way that we could prevent duplicates when importing config that already exists. perhaps we should keep this around but hide it from the general user a bit as it isn't really what people would want most of this time.

napiergit: I've been giving this some thought.  I don't think there is a need to persist the _id because all the collections have unique attributes, which inherently prevent duplicates.

rcrichton: Ok, but what we would loose is the ability to update existing collection unless you look them up by their unique properties. Are you sure this is possible for each collection type? If so, then maybe we can simplify an only have 1 export option which doesn't include the `_id`s.

cc/ @hnnesv for your thoughts.

hnnesv: Yeah +1 for simplifying and dropping the `_ids`. My feeling is that if the user is dealing with things at that level they should rather deal with mongo directly. Looks like each collection type has a unique key attribute, from a quick scan...

rcrichton:
@nthfloor2 thanks, this has started working since the last commit. I'm still getting a blank screen when ignoring all conflicts through. Could you check that out.

app/scripts/controllers/exportImport.js
```diff
@@ -28,29 +29,59 @@ angular.module('openhimConsoleApp')
...
+        resolve: {
+          data: function () {return $scope.validatedData;}
+        }
+      });
+
+      modalInstance.result.then(function(importResults) {
+        $scope.importStatus = 'done';
+        $scope.importResults = importResults;
+      }, function() {
+        console.info('Modal dismissed at: ' + new Date());
```
rcrichton: Lets remove this, I don't think we should be writing anything to the console unless there is an error.

app/scripts/controllers/exportImport.js
```diff
@@ -28,29 +29,59 @@ angular.module('openhimConsoleApp')
...
-    $scope.exportCollections.ContactGroups = ContactGroups;
-
-    // set inital reset ( default export option )
-    $scope.resetExportOptions();
-    
+    var exportObject = Api.Metadata.query(function(result) {
+      getMetadataSuccess(result);
+    }, function(err) {
+      getMetadataError(err);
+    });
```
rcrichton: Could you not simply do `var exportObject = Api.Metadata.query(getMetadataSuccess, getMetadataError)`?

app/scripts/controllers/exportImport.js
```diff
@@ -198,122 +213,36 @@ angular.module('openhimConsoleApp')
...
-            });
-          }else{
-            record.$save(insertParams, function(){
-              importSuccess();
-            }, function(err){
-              importFail(model, value, err);
-            });
-          }
+    var validateSuccess = function(result) {
+      console.log('succesfully validated uploaded file');
```
rcrichton: I think we can remove this to.

app/scripts/controllers/exportImport.js
```diff
@@ -198,122 +213,36 @@ angular.module('openhimConsoleApp')
...
+    };
 
-        });
+    $scope.validateImportFile = function(data) {
+      $scope.importStatus = 'progress';
 
+      Api.MetadataValidation.save(data, function(result){
+        validateSuccess(result);
+      }, function(err){
+        validateFail(err);
```
rcrichton: This can be simplified to `Api.MetadataValidation.save(data, validateSuccess, vallidateFail)`

app/scripts/controllers/exportImportModal.js
```diff
@@ -0,0 +1,182 @@
...
+              else if(item.model==='ContactGroups') {$scope.resolvedData.ContactGroups.push(item.record);}  
+          });
+
+          // read the import script data and process
+          if($scope.resolvedData.Channels.length > 0 ||
+            $scope.resolvedData.Clients.length > 0 ||
+            $scope.resolvedData.Users.length > 0 ||
+            $scope.resolvedData.Mediators.length > 0 ||
+            $scope.resolvedData.ContactGroups.length > 0) { $scope.runImportFile($scope.resolvedData, callback); }
+        }
```
rcrichton: Callback should be called even if nothing is imported.

app/scripts/controllers/exportImportModal.js
```diff
@@ -0,0 +1,182 @@
...
+
+    /***************************************************/
+    /**         Initial page load functions           **/
+    /***************************************************/
+
+    $scope.outcome = {
+      selectedAll: false
+    };
+    $scope.ngErrors = {};
+    $scope.conflicts = [];
```
rcrichton: I think by default every conflict should have an action of ignore so that the radio button is set.

app/views/exportImport.html
```diff
@@ -72,33 +64,41 @@
...
 
-              <div id="no-more-tables" ng-if="importFail">
-                <table class="table table-bordered" style="100%">
+            <img src="images/loading.gif" alt="Loading..." ng-if="importStatus == 'progress'"/>
+            
+            <label class="btn btn-primary" ng-click="showConflictModal()" ng-if="importStatus == 'resolvingConflicts'"><strong>Resolve Conflicts</strong></label>
+
+            <div ng-if="importStatus == 'done'">
+              <div>
+                <label class="successfulImports">Successfull Imports ( {{ numberOfSuccessfulImports() }} )</label>
```
rcrichton: I know this was inherited from before, but I don't think the look of change this to `Successful Imports ( 5 )` I think we can change this to `Successful Imports: 5` for this one and the others.

app/views/exportImportModal.html
```diff
@@ -0,0 +1,95 @@
...
+  <div class="modal-body">
+    <div ng-if="conflicts.length === 0" class="med-no-config">
+      <center>
+        <span>There are no conflicts with this import.</span>
+      </center>
+    </div>
+
+    <div id="no-more-tables" ng-if="importStatus == 'resolvingConflicts'">
+      <div>
+        <label class="successfulImports">Successfull Imports ( {{validImports.length}} )</label>
```
rcrichton: Again, let's make these prettier.

test/spec/controllers/exportimport.js
```diff
@@ -213,49 +204,4 @@ describe('Controller: ExportImportCtrl', function () {
     expect(scope.downloadLink.indexOf('blob:http://localhost:8080/')).to.be.above(-1);
   });
```
rcrichton: There are a number of function attached to the scope that can be tested for  this controller. I think at least the important ones should have test.

test/spec/controllers/exportimportModal.js
```diff
@@ -0,0 +1,191 @@
...
+    var spy = this.spy(modalInstance, 'close');
+
+    // lets change some data in our data object before doing the import
+    scope.conflicts.forEach(function(item){
+      item.action = 'overwrite';
+    });
+
+    // execute the import function
+    scope.saveImport(function() {
+      sinon.assert.calledOnce(spy);
```
rcrichton: I think we should also verify that a request got sent. This can be done with httpBackend.

test/spec/controllers/exportimportModal.js
```diff
@@ -0,0 +1,191 @@
...
+    var spy = this.spy(modalInstance, 'close');
+
+    // lets change some data in our data object before doing the import
+    scope.conflicts.forEach(function(item){
+      item.action = 'ignore';
+    });
+
+    // execute the import function
+    scope.saveImport(function() {
+      sinon.assert.notCalled(spy);
```
rcrichton: It seems to me like 'close' should still be called if everything is ignored. Else you get a blank screen.

rcrichton:
@nthfloor2 @napiergit Cool this is looking good. I have added a few comments above that need to be addressed but overall this is working well. Since both of you worked on this I'm not sure who is doing the changes but I guess it will be @nthfloor2 as he was working on it last?

nthfloor2:
@rcrichton changes have been made based on feedback. When you have a chance, it is ready for review again.

rcrichton:
@nthfloor2 LGTM :+1: Works great.

PR #401, opened 2016-08-01
Ohm 257 metrics graphs

hnnesv:
OHM-257
screen shot 2016-08-01 at 10 31 39
screen shot 2016-08-01 at 10 32 29


app/scripts/controllers/channelMonitoring.js
```diff
@@ -267,28 +135,35 @@ angular.module('openhimConsoleApp')
...
-        endDate: endDate
-      }, statusMetricsSuccess, statusMetricsError);
+    $scope.updateMetrics = function() {
+      Alerting.AlertReset('load');
+      Alerting.AlertReset('responseTime');
+      Alerting.AlertReset('status');
 
+      Metrics.refreshDatesForSelectedPeriod($scope.selectedDateType);
+      updateTimeseriesMetrics();
+      updateStatusMetrics();
```
rcrichton: Minor: Maybe call this channel metrics. We don't have the concept of status metrics anymore.

app/scripts/controllers/dashboard.js
```diff
@@ -181,128 +60,107 @@ angular.module('openhimConsoleApp')
...
 
-    };
+    }
 
-    $scope.statusMetricsSuccess = function(statusResults){
-      if ( statusResults.length === 0 ){
-        Alerting.AlertAddMsg('status', 'warning', 'There has been no transactions received for today');
-      }else{
-        $scope.updateStatusBarChart(statusResults);
+    function statusMetricsSuccess (metrics) {
```
rcrichton: Minor again: Channel metrics?

rcrichton:
@hnnesv this is looking really awesome! I've added a few very minor comments above.

I've also run into something else that we may want to consider. I by mistake ended up doing a custom query for the previous 5 years broken down by hour... It's currently still running as I type this and has locked up the entire UI with what seems to be js processing :) Maybe we should set some limits on what is reasonable. For example if the timeperiod is > 1y you can't breakdown by hour or by day, only by month. What do you think?

rcrichton:
Also, would you be able to add 'from' and 'to' labels in front of the date picker. When they already have values in the dates it's hard for the user to know which is which (although fairly obvious). I've also noticed an issue where the 'There has been no transactions received for the queried timeframe' warning is repeated a number of times when metrics first load. I haven't quite figured out when this occurs. I'l update soon.

rcrichton:
I can't seem to lock down when the warnings get repeated but it seems to occur most frequently when I click on a stack graph to go to a channel's metrics page and there aren't any metrics available for the default timeseries.

hnnesv:
Yeah fully agree on restricting the breakdown when the period is large; I noticed that too but forgot about it.

Will investigate the duplicate warnings - I haven't been able to reproduce it yet though :/

hnnesv:
I *think* I know what was causing the problem: I refactored the morris graph updating directive to empty the html and reload it, rather than just call the setData method. However I think this was causing all sorts of problems, including I noticed the bar chart links not behaving correctly.

rcrichton:
@hnnesv Ok cool, is it ready for re-review yet or are you still fixing other stuff?

hnnesv:
Oops, forgot to drag the ticket across. Yup ready.

rcrichton:
@hnnesv it seems when I run with grunt this I get some bower dependencies getting changed:

```diff
diff --git a/app/index.html b/app/index.html
index 48688c7..a5e9db5 100644
--- a/app/index.html
+++ b/app/index.html
@@ -136,8 +136,7 @@
     <script src="bower_components/d3/d3.js"></script>
     <script src="bower_components/angular-bootstrap/ui-bootstrap-tpls.js"></script>
     <script src="bower_components/angular-taglist/js/angular-taglist-directive.js"></script>
-    <script src="bower_components/eve/eve.js"></script>
-    <script src="bower_components/raphael/raphael.min.js"></script>
+    <script src="bower_components/raphael/raphael.js"></script>
     <script src="bower_components/mocha/mocha.js"></script>
     <script src="bower_components/morris.js/morris.js"></script>
     <script src="bower_components/moment/moment.js"></script>
```

hnnesv:
Maybe try removing your current bower_components and running from scratch? I think the one being replaced is the newer version.

app/scripts/directives/metrics-date-type-selector.js
```diff
@@ -13,6 +14,31 @@ angular.module('openhimConsoleApp')
...
+            scope.selected.from = scope.selected.until;
+          } else if (scope.selected.type === 'minute' && diff('minutes') > maxBreakdown) {
+            scope.selected.type = 'hour';
+          } else if (scope.selected.type === 'hour' && diff('hours') > maxBreakdown) {
+            scope.selected.type = 'day';
+          } else if (scope.selected.type === 'day' && diff('days') > maxBreakdown) {
+            scope.selected.type = 'month';
+          } else if (scope.selected.type === 'month' && diff('months') > maxBreakdown) {
+            scope.selected.type = 'year';
+          }
```
rcrichton: This only moves the disaggregation up one level, the next level can still be above 120. We need to scan until we find a breakdown that is < 120. Also, it may be better if the relevant options are disabled in the drop box as a visual indicator to the user about what is happening.

rcrichton:
@hnnesv thanks that sorted the bower thing. I though the bower install was updated automatically, obviously not.

app/scripts/directives/metrics-date-type-selector.js
rcrichton: You may be able to just change the `else if`s to `if`s.

hnnesv:
Good catch! Fixed the handling.

rcrichton:
Thanks @hnnesv LGTM, I did notice we are missing `week` in the breakdown options. Is this intentional? I'm happy to merge and add an issue for this if it's difficult to add now, but if it's quick it may be good to include. :+1:

hnnesv:
Aah oops, no I just missed it - nope, it's easy enough to add...

hnnesv:
Done

rcrichton:
:+1:

Issue #359, opened 2016-02-05
Grunt Build Bug - imagemin


BMartinos:
When running "grunt" or "grunt build" the build fails.

"Warning: Running "imagemin:dist" (imagemin) task - Fatal error: Cannot read property 'contents' of undefined" 
![grunt-error](https://cloud.githubusercontent.com/assets/7311421/12846840/97cd654c-cc17-11e5-9474-e5278a86aad7.png)

BMartinos:
Fixed in PR #406

Issue #400, opened 2016-07-29
OpenHIM Admin Console - Audit Log - Page Redirects When Alphanumeric


rcrichton:
See https://github.com/pepfar-datim/OHIE/issues/297

BMartinos:
Fixed in PR #406

PR #406, opened 2016-08-10
Ohm 291 audit log redirect bug

BMartinos:
This PR relates to issue: #400 

There was a bug with how the location redirect URLs were constructed. When the Console was run in sub-domain, the location URL wouldn't be valid and redirect the user to the web server landing page. Same result were experienced in Apache.

A dependency upgrade also fixed bug: #359

### DIFF
--- a/app/scripts/controllers/audits.js
+++ b/app/scripts/controllers/audits.js
@@ -322,13 +322,15 @@ angular.module('openhimConsoleApp')
     $scope.applyFiltersToUrl = function(){
       var curHashParams = window.location.hash;
       var filters = $scope.returnFilters('urlParams');
-      var newHash = '#/audits?'+filters.substring(1);
+      var newHash = '/audits?'+filters.substring(1);
 
       // just refresh audits list - no new params
       if ( curHashParams ===  newHash ){
         $scope.refreshAuditsList();
       }else{
-        var baseUrl = $location.protocol() + '://' + $location.host() + ':' + $location.port() + '/';
+        var absUrl = $location.absUrl();
+        var absUrlPath = $location.url();
+        var baseUrl = absUrl.replace(absUrlPath, '');
         window.location = baseUrl + newHash;
       }
     };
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/audits.js
```diff
@@ -407,12 +411,20 @@ angular.module('openhimConsoleApp')
...
-      $scope.refreshAuditsList();
+      // clear all filter parameters
+      $location.search({});
+
+      // get the filter params object after clearing them
+      var filterParamsAfterClear = JSON.stringify( angular.copy( $location.search() ) );
+
+      // if the filters object stays the same then call refresh function
+      // if filters object not the same then angular changes route and loads controller ( refresh )
+      if ( filterParamsBeforeClear === filterParamsAfterClear ){
```
rcrichton: I think you can skip converting these to string and rather use `angular.equals` to do this check directly on the objects.

rcrichton:
@BMartinos this looks good, nice catch. I'm still getting some redirects when entering invalid characters in the some of the date fields. Perhaps we should change this to work `onBlur` rather than right which seems to be `onChange`.

app/scripts/controllers/audits.js
BMartinos: Great works well!

BMartinos:
Will have a look at whats causing the dates to redirect as well.

BMartinos:
Hey @rcrichton Ive made the change as suggested and works well. regarding the change for the date, unfortunately ng-blur wouldn't have worked as its not triggered when using the mouse to click on a specific date. This then only works when typing the date. Ive added a function to check if the date is valid and only then run the apply filters function.

app/scripts/controllers/audits.js
```diff
@@ -412,17 +419,19 @@ angular.module('openhimConsoleApp')
       $scope.settings.list.tabview = 'same';
 
       // get the filter params object before clearing them
-      var filterParamsBeforeClear = JSON.stringify( angular.copy( $location.search() ) );
+      //var filterParamsBeforeClear = JSON.stringify( angular.copy( $location.search() ) );
```
rcrichton: Remove commented out code.

app/scripts/controllers/audits.js
```diff
@@ -412,17 +419,19 @@ angular.module('openhimConsoleApp')
...
+      var filterParamsBeforeClear = angular.copy( $location.search() );
 
       // clear all filter parameters
       $location.search({});
 
       // get the filter params object after clearing them
-      var filterParamsAfterClear = JSON.stringify( angular.copy( $location.search() ) );
+      // var filterParamsAfterClear = JSON.stringify( angular.copy( $location.search() ) );
```
rcrichton: remove commented out code

rcrichton:
LGTM :+1:

rcrichton:
Just remove the commented out code.

PR #407, opened 2016-08-10
OHM-296 list visualizers view

rcrichton:


app/scripts/controllers/visualizer.js
```diff
@@ -4,25 +4,15 @@
 angular.module('openhimConsoleApp')
   .controller('VisualizerCtrl', function ($scope, $http, $interval, $window, login, Api, Alerting, Fullscreen) {
 
+    // initialize global variables
+    var settingsStore = {}; // a place the push current settings when switching to fullscreen
+    var visualizerUpdateInterval, updatePeriod, diffTime, lastUpdate, maxSpeed, pad;
+
     $scope.loadingVisualizer = true;
     $scope.loadingVisualizerError = false;
     $scope.loadingVisualizerErrorMsgs = [];
```
BMartinos: The above two don't seem to be used anymore, but maybe they are kept in for future use?

app/views/visualizer.html
```diff
@@ -10,48 +9,72 @@
           <div class="panel-title">
             <h2><i class="glyphicon glyphicon-road"></i>&nbsp;&nbsp;Visualizer</h2>
             <p class="text-muted">
-              The visualizer displays a live view of how transactions are being routed through the OpenHIM.
+              The visualizer displays a live view of how transactions are being routed through the OpenHIM. Multiple visualizers can be created and these are shared amoung admin users. Pick a visualizer on the left to start viewing or create a new one.
```
BMartinos: Small typo: amoung

BMartinos:
Hey @rcrichton The code looks and works great for me! Just some minor comments above. 

I am able to switch between Visualiser views easily and didn't encounter any issues. There are some styling changes we can make but not sure if its important to focus on those within this PR. 

The only other issue ive encountered, which might have been there for a while already is Firefox not rendering the Visualiser correctly. Screenshot below:
screenshot from 2016-08-11 09 50 43

app/scripts/controllers/visualizer.js
rcrichton: I believe they are still used, in line 52, 54, 66, 68 etc.

rcrichton:
@BMartinos hmm interesting that firefox doesn't like the visualizer. I haven't touched the rendering code so I think this is an existing issue. Could you add a jira issue for this?

BMartinos:
Just some of the styling comments i have:

1. List items border-radius should be removed to keep consistency with main styles
2. Reduce padding slightly to save some space
3. Move the "+ Visualiser" button to the top of the list so its easy to add a new ones
4. When the screen moves smaller ( responsive - 1300px ) the list moves above the Visualiser. This might need to be adapted into a show/hide or popup feature. thinking in terms of having 20+ visualisers and not being able to see it at the bottom of the list.

These styling comments could be tackled in a separate issue if they take to long as it doesnt affect the functionality of the View/Controller

app/scripts/controllers/visualizer.js
BMartinos: Ahhh yes you right, My search didn't pick them up sorry.

rcrichton:
@BMartinos ok cool, I've adjusted the padding and the border-radius, thanks. I tried the '+ Visualizers' button at the top and it just looks weird. Maybe lets leave it as is for now and if it becomes a problem we can figure something out. Also, for the responsive stuff, I don't think there is a big use case for viewing the visualizer on mobile. I think the responsiveness is fine (nothing get displayed weird) and I don't want to add a bunch off new styling for something that won't really be used on mobile. Do you agree?

BMartinos:
Great thanks @rcrichton !

I agree that we can leave out those responsive changes for now as its not important and wont really be used on mobile, We can also leave the "+ Visualisers" button where it is until we notice it becoming a problem. 

Just a very super tiny style for the nav-pill when hovering
.nav-pills > li > a:hover {
    border-radius: 0;
}

Everything works well so this is good to merge :+1:

rcrichton:
@BMartinos cool, added that.

PR #413, opened 2016-08-12
OHM 306 - implement add new visualizer logic

nthfloor2:


rcrichton:
@nthfloor2 it seems like this PR is showing changes for a bunch of new issues. Perhaps `ohm-292` needs to be merged with `master` to make this PR show up correctly. I'l try review what a can for now, but it's a bit hard to see the changes.

app/scripts/controllers/visualizerModal.js
```diff
@@ -0,0 +1,204 @@
...
+    if (settingsStore) {
+      $scope.update = true;
+      
+      
+    }else{
+      $scope.update = false;
+
+      // create visualizer settings properties
+      $scope.settings = {};
+      $scope.settings.visualizer = {};
```
rcrichton: I think that we should drop to just using `$scope.visualizer` otherwise it becomes a bit confusing. This just be a rename really.

app/scripts/controllers/visualizerModal.js
```diff
@@ -0,0 +1,204 @@
...
+
+    $scope.addSelectedMediator = function(){
+      $scope.settings.visualizer.mediators.push({ mediator: $scope.visualizer.addSelectMediator.urn, name: $scope.visualizer.addSelectMediator.name, display: $scope.visualizer.addSelectMediator.name });
+      $scope.visualizer.addSelectMediator = null;
+    };
+
+    $scope.addComponent = function(){
+      if( $scope.visualizer.addComponent.eventType ){
+        $scope.settings.visualizer.components.push({ eventType: $scope.visualizer.addComponent.eventType, eventName: $scope.visualizer.addComponent.eventName, display: $scope.visualizer.addComponent.display });
+        $scope.visualizer.addComponent.eventType = '';
```
rcrichton: To support my previous comment these would then need to drop the `.visualizer` prefix.

app/scripts/controllers/visualizerModal.js
```diff
@@ -0,0 +1,204 @@
...
+          $scope.ngError.hasErrors = true; 
+        }
+      }      
+
+      if(settings.components === undefined || settings.components.length === 0){
+        $scope.ngError.hasErrors = true;
+        $scope.ngError.hasNoComponents = true;
+      }
+
+      if(settings.mediators === undefined || settings.mediators.length === 0){
```
rcrichton: Mediator shouldn't be a required field, they are optional.

app/scripts/controllers/visualizerModal.js
```diff
@@ -0,0 +1,204 @@
...
+        Alerting.AlertAddMsg('hasErrors', 'danger', 'There are errors on the form.');
+        callback('Error: Form has errors');
+      } else {
+        callback();
+      }
+    };
+
+    var notifyUser = function(){
+      // reset backing object and refresh users list
+      Notify.notify('channelsChanged');
```
rcrichton: This is probably from existing code, but `channelChanged` seems to wrong. Is this even used in the visualizer list controller?

rcrichton:
I also notice that the time components off the visualizer don't get added properly, this may be because of a previous bug in the model. An update with master on this branch may fix this.

app/scripts/controllers/visualizerModal.js
nthfloor2: `$scope.visualizer` is acting as a sort of viewModel due to the fact that we don't store the entire channel and mediator config(s) in the visualizer settings.

nthfloor2: Oh wait never mind I see what you mean - we just remove the 'visualizer' part but still keep the viewModel.

rcrichton:
Err, ignore the previous comment as this would be a core issue... I'm testing with core `ohm-292` branch which doesn't seems to work, which should I use?

BMartinos:
I think this might be due to that we've changed core to use Id instead of the Visualizer name

app/scripts/controllers/visualizer.js
```diff
@@ -195,6 +195,21 @@ angular.module('openhimConsoleApp')
...
 
+    // function to add new visualizer to list
+    $scope.addVisualiser = function addVisualiser(){
+
+      // open visualizer settings modal 
+      $modal.open({
+        templateUrl: 'views/visualizerModal.html',
+        controller: 'VisualizerModalCtrl',
+        resolve: {
+          settingsStore: function () { return null; },
```
rcrichton: I don't think this get used. Does it need to be removed?

app/scripts/controllers/visualizerModal.js
rcrichton: It looks like we store mediators and channels in `$scope.mediators` and `$scope.channels` is there something I'm missing here?

app/scripts/controllers/visualizerModal.js
rcrichton: Btw, this is how a refresh of the channels list is triggered on an update of a channel.

app/scripts/services/rest.js
```diff
@@ -91,7 +91,10 @@ angular.module('openhimConsoleApp')
       }),
 
       // Visualizer API
-      Visualizers: $resource(server + '/visualizers/:name', { name: '@name' }),
+      Visualizers: $resource(server + '/visualizers/:name', { name: '@name' }, {
+        save: { method: 'POST' },
+        query: { method: 'GET', isArray: true }
```
rcrichton: I think these are the defaults of save and query, so this object can be removed.

app/views/partials/visualizer-settings.html
```diff
@@ -0,0 +1,310 @@
+
+<h4>Visualizer Settings</h4>
+
+<div class="panel panel-default">
+  <!-- <div class="panel-heading">General Settings</div> -->
```
rcrichton: Delete commented out code.

app/views/partials/visualizer-settings.html
```diff
@@ -0,0 +1,310 @@
...
+            </tbody>
+          </table>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+
+
+<div class="panel-group" id="accordion">
```
rcrichton: Perhaps add a heading of 'Advanced settings' here to let the user know they don't  need to touch these.

app/views/visualizerModal.html
```diff
@@ -0,0 +1,35 @@
...
+<div class="modal-footer">
+  <div>
+    <!-- hasErrors Alerts -->
+    <alert style="display: inline-block; margin: 0px 0px 0px 5px; padding: 8px;" ng-repeat="alert in alerts.hasErrors" type="alert.type">
+      <i class="glyphicon glyphicon-warning-sign"></i>{{alert.msg}}
+    </alert>
+    <!-- hasErrors Alerts -->
+    <button type="button" class="btn btn-default" ng-click="cancel()">Close</button>
+    
+    <button type="button" class="btn btn-primary" ng-click="submitSettingsForm()">
```
rcrichton: `submitSettingForm` seems like a weird name from the old code, maybe change this to `saveVisualizer` or something more descriptive.

app/views/visualizerModal.html
```diff
@@ -0,0 +1,35 @@
...
+      <span ng-if="update == true">Save changes</span>
+      <span ng-if="update == false">Create Visualizer</span>
```
rcrichton: I've noticed some issues when update isn't set yet and no text is displayed. Maybe change this just to `ng-if="update"` and `ng-if="!update"`, then I think it will work with null and undefined to.

test/spec/controllers/visualizerModal.js
```diff
@@ -0,0 +1,306 @@
...
+  beforeEach(module('openhimConsoleApp'));
+
+  // setup config constant to be used for API server details
+  beforeEach(function(){
+    module('openhimConsoleApp', function($provide){
+      $provide.constant('config', { 'protocol': 'https', 'host': 'localhost', 'port': 8080, 'title': 'Title', 'footerTitle': 'FooterTitle', 'footerPoweredBy': 'FooterPoweredBy' });
+    });
+  });
+
+  // defaultVisualizerSettings  = readJSON('app/config/visualizer.json');
```
rcrichton: delete commented out code

rcrichton:
@nthfloor2 I've added some comment above. Overall it's looking good and I like the new partial and tests.

app/scripts/controllers/visualizer.js
nthfloor2: I left it there to be used while editing to pass the selected visualizer data through to the modal. Are you doing that some other way?

rcrichton: I've added new properties for that, so maybe this one can be removed. Wasn't quite sure how this should have been used when I was doing it.

app/scripts/controllers/visualizerModal.js
```diff
@@ -12,22 +12,16 @@ angular.module('openhimConsoleApp')
...
-    }else{
-      $scope.update = false;
+    $scope.update = false;
 
-      // create visualizer settings properties
-      $scope.settings = {};
-      $scope.settings.visualizer = {};
+    // create visualizer settings properties
+    $scope.settings = {};
+    $scope.settings.visualizer = {};
```
rcrichton: Maybe I'm just being picky, but can't we rather store the visualizer setting in `$scope.visualizer` there is no need for the extra `.settings` object. It just makes the code look a bit cleaner.

app/scripts/controllers/visualizerModal.js
```diff
@@ -136,11 +130,6 @@ angular.module('openhimConsoleApp')
         $scope.ngError.hasNoComponents = true;
       }
 
-      if(settings.mediators === undefined || settings.mediators.length === 0){
-        $scope.ngError.hasErrors = true;
-        $scope.ngError.hasNoMediators = true;
-      }
-
       if(settings.channels === undefined || settings.channels.length === 0){
```
rcrichton: Could the settings parameter of this function be renamed visualizer or vis. Setting doesn't seems to make sense here.

rcrichton:
Looking good @nthfloor2 Just a few more minor comments, I think we have inherited some naming fluff from the previous code and this is a good time to try resolve this.

rcrichton:
LGTM :+1:

PR #416, opened 2016-08-19
OHM-309 edit and clone visualizers

rcrichton:


hnnesv:
@rcrichton overall it's working well! The only issue I'm seeing form testing is that the main visualizers page doesn't refresh after editing; it displays the old visualizer until a manual page refresh.

Starting with the code review itself now...

app/scripts/controllers/visualizerModal.js
```diff
@@ -1,29 +1,37 @@
...
-    $scope.visualizer = {};
-
-    // load default visualizer config for new user
-    $http.get('config/visualizer.json').success(function( visualizerConfig ) {
-      angular.extend( $scope.visualizer, angular.copy( visualizerConfig ) );
-    });
+    if (visualizer) {
+      $scope.update = true;
+      // make a copy of the object so that the original doesn't get changed until we save
+      $scope.visualizer = JSON.parse(angular.toJson(visualizer));
```
hnnesv: Any particular reason to copy like this instead of `angular.copy` like below? Either way's fine; I'm just curious...

hnnesv:
Looks good! - just the refreshing issue

app/scripts/controllers/visualizerModal.js
rcrichton: I was getting call stack exceeded with `angular.copy`, I think now that I'm using `angular.toJson` it may work again as it cuts out all the angular specific properties (which seem to have recursive references). I'l give this another go.

rcrichton:
@hnnesv cool, I'l have a look at the refresh issue. I was pretty sure this was working, but I may have missed it.

app/scripts/controllers/visualizerModal.js
rcrichton: Same problem still exists, I can't find a better way to do this.

hnnesv: Cool no worries, not a big deal

hnnesv:
👍 Works great!

app/scripts/controllers/visualizerModal.js
rcrichton: Future version of angular added a `toJSON()` method to resource object but the version we are using doesn't have this yet.

hnnesv: Aah, cool

PR #419, opened 2016-09-07
OHM-337: Add functionality to persist transaction filters to user session

napiergit:


de-laz:
@napiergit Looks good to me :+1:

PR #421, opened 2016-09-12
Ohm 354 filter by time

napiergit:
Before merging, wait for PR #420, to be merged.

### DIFF
--- a/app/scripts/controllers/transactions.js
+++ b/app/scripts/controllers/transactions.js
@@ -254,12 +254,12 @@ angular.module('openhimConsoleApp')
   
       if(filterDateStart || filterDateEnd) {
         var buildDateFilterObject = function(object, comparator, date) {
-          object[comparator] = moment(date).format();
+          object[comparator] = moment(new Date(date)).format();
         };
         var dateFilterObject = {};
         
         if(filterDateStart) { buildDateFilterObject(dateFilterObject, '$gte', filterDateStart); }
-        if(filterDateEnd) { buildDateFilterObject(dateFilterObject, '$lte', moment(filterDateEnd).endOf('day')); }
+        if(filterDateEnd) { buildDateFilterObject(dateFilterObject, '$lte', filterDateEnd); }
         
         filtersObject.filters['request.timestamp'] = JSON.stringify( dateFilterObject );
       }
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/transactions.js
```diff
@@ -251,10 +252,17 @@ angular.module('openhimConsoleApp')
       // date filter
       filterDateStart = $scope.settings.filter.startDate;
       filterDateEnd = $scope.settings.filter.endDate;
-      if(filterDateStart && filterDateEnd) {
-        var startDate = moment(filterDateStart).format();
-        var endDate = moment(filterDateEnd).endOf('day').format();
-        filtersObject.filters['request.timestamp'] = JSON.stringify( { '$gte': startDate, '$lte': endDate } );
+  
+      if(filterDateStart || filterDateEnd) {
+        var buildDateFilterObject = function(object, comparator, date) {
```
BMartinos: Maybe move this function of the parent function. It gets created every time the 'returnFilters' function is run.

app/scripts/controllers/transactions.js
```diff
@@ -577,6 +594,13 @@ angular.module('openhimConsoleApp')
 
       // execute refresh if no errors
       if($scope.ngError.hasErrors === false) {
+        
+        // Set dateApplied flag if a date filter is being applied
+        if($scope.settings.filter.startDate || $scope.settings.filter.endDate) {
+          $scope.dateApplied = true;
+        } else {
+          $scope.dateApplied = false;
+        }
```
BMartinos: It doesn't look like this code has any value within this function. This check can be made directly in the filtersApplied function (ln: 695)

BMartinos:
Hey @napiergit Its look great! Just have a few minor comments above. Then one other thing i noted was that if i have both date filters applied, and i manually remove one of them, it doesn't trigger the reload. only when i select dates does it trigger.

app/scripts/controllers/transactions.js
napiergit: The date filter is only triggered on blur, but the scope variable is set on date selected, resulting in the (Filters Applied) notification being displayed before the filter is applied.

BMartinos:
Great thank @napiergit Everything looks great. Just have a look at the seconds not applying constantly then we should be good to go.

BMartinos:
Thanks @napiergit this is working great! :+1:

PR #422, opened 2016-09-13
OHM-355: Add client name to transaction list

napiergit:
fixes https://github.com/jembi/openhim-core-js/issues/699

### DIFF
--- a/app/scripts/controllers/transactions.js
+++ b/app/scripts/controllers/transactions.js
@@ -231,7 +231,7 @@ angular.module('openhimConsoleApp')
       $scope.clients.map(function(client) {
         clientsIdNameMap[client._id] = client.name;
       });
-    }
+    };
     // clients used for advanced filter dropdown and client name on transaction list
     Api.Clients.query(clientsSuccess);
 
--- a/app/scripts/controllers/transactions.js
+++ b/app/scripts/controllers/transactions.js
@@ -221,9 +221,19 @@ angular.module('openhimConsoleApp')
         }
       });
     }, function(){ /* server error - could not connect to API to get channels */ });
-
-    // clients used for advanced filter dropdown
-    $scope.clients = Api.Clients.query();
+    
+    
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/transactions.js
```diff
@@ -221,9 +221,19 @@ angular.module('openhimConsoleApp')
...
-    $scope.clients = Api.Clients.query();
+    
+    
+    var clientsIdNameMap = [];
+    
+    var clientsSuccess = function(clients) {
+      $scope.clients = clients;
+      // Create client map of id to name
+      $scope.clients.map(function(client) {
+        clientsIdNameMap[client._id] = client.name;
```
BMartinos: Might be better to make this a scope variable, Then we would need the below function to be called for each client lookup. The the scope var can be used in the html instead of calling a function

app/scripts/controllers/transactions.js
```diff
@@ -749,6 +759,11 @@ angular.module('openhimConsoleApp')
         updateUser(currentUser);
       });
     };
+    
+    // Get client name from client id
+    $scope.getClientNameFromClientID = function(clientId) {
```
BMartinos: Following from the above comment, this will no longer be needed

app/views/transactions.html
```diff
@@ -103,6 +103,7 @@ <h3 style="margin: 20px;" ng-if="transactionsSelected.length < defaultBulkRerunL
                   <th>Channel</th>
                   <th>Status</th>
                   <th>Request Time</th>
+                  <th>Client</th>
```
BMartinos: Having the client on the end looks a little bit out of place. Lets keep the request time at the end and move the client next to the 'Channel' column

BMartinos:
Looks good so far @napiergit! just a few minor comments.

BMartinos:
Looks great! Thanks @napiergit :+1:

PR #427, opened 2016-12-14
OHM-387 Fix response time labels

bausmeier:
Update the graphs on the dashboard and channel metrics pages to show the
label for the response time as "Reponse Time" instead of "Transactions".


rcrichton:
@bausmeier looks good but the tests are failing. I think you need to update some expectations.

bausmeier:
Thanks @rchrichton. I didn't realise that there were tests for this. They were asserting on the wrong value in the first place 🙈

rcrichton:
Haha, cool looks good :+1:

PR #429, opened 2017-02-10
OHM-382: added alert/warning when exporting keystore

kaweesi:


app/index.html
```diff
@@ -140,7 +140,7 @@
     <script src="bower_components/d3/d3.js"></script>
     <script src="bower_components/angular-bootstrap/ui-bootstrap-tpls.js"></script>
     <script src="bower_components/angular-taglist/js/angular-taglist-directive.js"></script>
-    <script src="bower_components/eve/eve.js"></script>
+    <script src="bower_components/eve-raphael/eve.js"></script>
```
kaweesi: this was auto edited, any explanation for this @rcrichton, @hnnesv !!!

app/scripts/controllers/exportImport.js
```diff
@@ -32,7 +32,7 @@ angular.module('openhimConsoleApp')
       $scope.showRecordOptions.ContactGroups = false;
       $scope.showRecordOptions.Keystore = false;
     };
-
```
kaweesi: these auto file formattings happened and i see no problems keeping them

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,11 @@ angular.module('openhimConsoleApp')
       var exportData = angular.copy( $scope.selectedExports );
       var textFile = null;
 
+      // OHM-382: warn when exporting key
+      if(exportData.Keystore.length > 0) {
```
kaweesi: this is the real required change by this issue

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,11 @@ angular.module('openhimConsoleApp')
...
+        alert("The key will be exported and should be protected");
```
kaweesi: i wonder what could be the easiest way to manage locations within this node js framework if we in the future translated openHIM into other languages!

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,11 @@ angular.module('openhimConsoleApp')
       var exportData = angular.copy( $scope.selectedExports );
       var textFile = null;
 
+      // OHM-382: warn when exporting key
+      if(exportData !== undefined && exportData.Keystore !== undefined && exportData.Keystore.length > 0) {
```
napiergit: You don't need the !== undefined in here.

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,11 @@ angular.module('openhimConsoleApp')
...
+        Alerting.AlertAddMsg('bottom', 'warning', 'the key will be exported and should be protected');
```
napiergit: Add this line above to reset existing alerts: Alerting.AlertReset();

napiergit: Also I think make it a top alert, this will involve fixing a bug in the view too. (in views/exportImport.html change line 17 to alert in alerts.top)

napiergit: Maybe make the description more detailed: 'Warning: The server's TLS private key will be exported and should be protected!'

app/scripts/controllers/exportImport.js
kaweesi: i fell into into some NPE here that's why i had to do this wrapping

app/scripts/controllers/exportImport.js
kaweesi: can you clarify this part; "this will involve fixing a bug in the view too. (in views/exportImport.html change line 17 to alert in alerts.top)"

.gitignore
```diff
@@ -3,6 +3,7 @@ dist
 .tmp
 .sass-cache
 .idea/**
+.project
```
kaweesi: i am using eclipse IDE and so i have to ignore its .project directory

app/scripts/controllers/exportImport.js
napiergit: I see in the view there is no alerts.top section.  It is called alerts.role for some reason.  That is a bug

app/scripts/controllers/exportImport.js
napiergit: That is very strange, I don't think the !== undefined makes any difference though.

app/scripts/controllers/exportImport.js
kaweesi: still not clear unfortunately to me

app/scripts/controllers/exportImport.js
kaweesi: o yeah, but i experienced it, will remove it and test once again to trace its cause

app/scripts/controllers/exportImport.js
bausmeier: The only problem with autoformatting is that it makes the relevant changes harder to review since there are unrelated changes in the diff. If you are going to make formatting changes then they should be done in a separate commit so that they can be reviewed independently of the actual change.

app/scripts/controllers/exportImport.js
napiergit: It will become clear if you try to make it a 'top' alert.

.gitignore
napiergit: I guess it's fine then

app/scripts/controllers/exportImport.js
kaweesi: o yeah, fixed at: ec05a74

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,12 @@ angular.module('openhimConsoleApp')
       var exportData = angular.copy( $scope.selectedExports );
       var textFile = null;
 
+      // OHM-382: warn when exporting key
+      Alerting.AlertReset();
```
napiergit: This can go in the if statement

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,12 @@ angular.module('openhimConsoleApp')
...
+      if(exportData !== undefined && exportData.Keystore !== undefined && exportData.Keystore.length > 0) {
+    	Alerting.AlertAddMsg('top', 'warning', 'Warning: The server\'s TLS private key will be exported and should be protected!');
```
napiergit: This indentation is not quite right

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,12 @@ angular.module('openhimConsoleApp')
       var exportData = angular.copy( $scope.selectedExports );
       var textFile = null;
 
+      // OHM-382: warn when exporting key
+      Alerting.AlertReset();
+      if(exportData !== undefined && exportData.Keystore !== undefined && exportData.Keystore.length > 0) {
```
napiergit: you can leave out the !== undefined

napiergit:
Hi @kaweesi there are still some changes that have not been made

app/scripts/controllers/exportImport.js
kaweesi: do you mean both of the undefined checks!! or one of them?

napiergit: both

app/scripts/controllers/exportImport.js
kaweesi: if the user exported keystore inclusive, the alert would be posted but without reloading page if they re-eported excluding keystore i needed to remove the alert that's why i put it out!! do you still want it enclosed within if statement?

app/scripts/controllers/exportImport.js
kaweesi: let me address this after @napiergit confirms the rest of my reviews here so that i can have one commit instead to have this all resolved!

app/scripts/controllers/exportImport.js
kaweesi: o, i think i should have made it !== null, even this has no impact! let re-run my HIM instance to check if i fall into the issue again that made me include this check

app/scripts/controllers/exportImport.js
napiergit: Good point.  This is fine as is then.

app/scripts/controllers/exportImport.js
napiergit: I would do this:
if(exportData && exportData.Keystore && exportData.Keystore.length > 0)

kaweesi:
all should be now resolved @napiergit !!!

app/scripts/controllers/exportImport.js
```diff
@@ -126,6 +126,12 @@ angular.module('openhimConsoleApp')
       var exportData = angular.copy( $scope.selectedExports );
       var textFile = null;
 
+      // OHM-382: warn when exporting key
```
bausmeier: There's no need for this comment. The related ticket can be found from the Git history if required.

kaweesi:
thanks  @napiergit and @bausmeier  for the reviews, i have addressed all your comments and squashed all my commits into one 👍

PR #430, opened 2017-02-10
OHM-310: refactored True/False to Yes/No

kaweesi:


napiergit:
Nice one @kaweesi LGTM :+1:

PR #431, opened 2017-02-13
OHM-416: Resolving TypeError: Cannot read property 'forEach' of null

kaweesi:


app/scripts/controllers/transactions.js
```diff
@@ -962,17 +962,22 @@ angular.module('openhimConsoleApp')
...
-          Api.Transactions.get({ transactionId: trx._id, filterRepresentation: 'simple' }, function(result) {
-            $scope.transactions.forEach(function(scopeTrx) {
-              if (scopeTrx._id === result._id) {
-                scopeTrx.status = result.status;
-              }
-            });
-          });
-        }
-      });
+      if ($scope.transactions !== null) {
```
napiergit: Hi @kaweesi nice one on finding the bug.

Instead of wrapping the whole function in an if statement, you could do something like: 
if(!$scope.transactions) { return } at the top of the function.

When testing locally I also found the function $scope.pollForLatest to be writing errors to console when the $scope.transactions was not ready, so can you please add that line to the top of that function too?

Also, double check the indentation.

napiergit:
@kaweesi can you just check the indentation and get the tests to pass (just some missing semi colons)

kaweesi:
i actually had them in the first place but removed them testing how smart this js framework would coverup

napiergit:
@kaweesi can you please change the indentation?

PR #436, opened 2017-08-25
Reverse transactions before adding them to the log

bausmeier:
The new transactions which are returned when polling for updates are
received in reverse chronological order from the core. Iterating through
them to add them to the current list of transactions reverses this
order. Since the API does not support a parameter to change the sort
order the new transactions need to be reversed before they are added to
the log.

OHM-460

### DIFF
--- a/app/scripts/controllers/transactions.js
+++ b/app/scripts/controllers/transactions.js
@@ -956,7 +956,7 @@ angular.module('openhimConsoleApp')
 
         Api.Transactions.query(filters, function(transactions) {
           lastPollingCompleted = true;
-          transactions.forEach(function(trx) {
+          transactions.reverse().forEach(function(trx) {
             $scope.transactions.unshift(trx);
             $scope.baseIndex--;
           });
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/transactions.js
```diff
@@ -956,7 +956,7 @@ angular.module('openhimConsoleApp')
 
         Api.Transactions.query(filters, function(transactions) {
           lastPollingCompleted = true;
-          transactions.forEach(function(trx) {
+          transactions.reverse().forEach(function(trx) {
```
rcrichton: I wonder if we should reverse the transactions or rather sort by date. Not sure we will always get the transaction back in this order if we refactor unless we build in a guarantee for this in the API.

bausmeier: The transactions shown in the log on initial load are also dependent on the order that the API returns them in. I guess the correct solution would be to add a sort query parameter to the endpoint and handle it in the core.

rcrichton: Yeah ok, I agree. If that's the case we should add a guarantee to the API.

rcrichton: Ah, I see we already do this: https://github.com/jembi/openhim-core-js/blob/master/src/api/transactions.coffee#L195

Ignore me :)

bausmeier: Oh, yes I knew that was there but wasn't sure what you meant by adding a guarantee 👍

PR #454, opened 2017-11-13
Ohm 514 read only channels

MattyJ007:


app/scripts/controllers/channelsmodal.js
```diff
@@ -1,5 +1,17 @@
 import { getTimeForTimezone, getTimezoneOffset } from '../utils'
 
+let defaultHttpMethods = {
```
bausmeier: Use `const` here since `defaultHttpMethods` shouldn't be reassigned.

app/scripts/controllers/channelsmodal.js
```diff
@@ -344,6 +387,21 @@ export function ChannelsModalCtrl ($scope, $uibModalInstance, $timeout, Api, Not
   /***************************************************************************/
 }
 
+function populateHttpMethods (channel, methods) {
+  if (channel.methods) {
+    for (let channelMethod of channel.methods) {
```
bausmeier: Use `const` here since `channelMethod` shouldn't be reassigned.

app/views/partials/channels-tab-basic-info.html
```diff
@@ -22,6 +22,40 @@
...
+    </div>
+    <div class="checkbox">
+        <label><input type="checkbox" ng-model="methods.TRACE" ng-checked="methods.TRACE">TRACE</label>
+    </div>
+    <div class="checkbox">
+        <label><input type="checkbox" ng-model="methods.CONNECT" ng-checked="methods.CONNECT">CONNECT</label>
+    </div>
+    <div class="checkbox">
+        <label><input type="checkbox" ng-model="methods.PATCH" ng-checked="methods.PATCH">PATCH</label>
+    </div>
```
bausmeier: Rather than having a chunk of HTML for every one of the HTTP methods we could iterate through the methods on `defaultHttpMethods` and create a checkbox for each one. That way we also only have a single place to maintain the list of methods.

app/views/partials/channels-tab-basic-info.html
```diff
@@ -22,6 +22,40 @@
   </div>
 </div>
 
+<div class="form-group-2columns" ng-show="channel.type == 'http'">
+    <div class="form-group" style="float: left;" ng-class="{ 'has-error' : ngError.methods }">
+    <label><span>*</span>Allowed HTTP methods:</label>
+    <div class="checkbox">
+        <label><input type="checkbox" ng-model="methods.POST" ng-checked="methods.POST">POST</label>
```
bausmeier: `ng-model` and `ng-checked` should not be used together:

> Note that this directive should not be used together with ngModel, as this can lead to unexpected behavior. - https://docs.angularjs.org/api/ng/directive/ngChecked

It should be one or the other, and by looking at the docs it seems like `ng-model` is more correct in this case.

app/scripts/controllers/channelsmodal.js
```diff
@@ -325,10 +357,21 @@ export function ChannelsModalCtrl ($scope, $uibModalInstance, $timeout, Api, Not
     }
   }
 
+  $scope.addHttpMethodsToChannel = function () {
+    $scope.channel.methods = []
+
+    Object.keys($scope.methods).forEach(function (key) {
```
bausmeier: No biggie, but it might be a little bit neater to use `filter` instead of `forEach`.

test/spec/controllers/channelsmodal.js
```diff
@@ -254,6 +254,8 @@ describe('Controller: ChannelsModalCtrl', function () {
     scope.channel.$update.should.have.been.called()
 
     scope.channel.should.have.property('name', 'ChannelName')
+    scope.channel.should.have.property('methods')
+    scope.channel.methods.length.should.equal(9)
```
bausmeier: I think that the tests can be a bit more comprehensive than this.

app/scripts/controllers/channelsmodal.js
```diff
@@ -51,6 +63,9 @@ export function ChannelsModalCtrl ($scope, $uibModalInstance, $timeout, Api, Not
   $scope.autoRetry = {}
   $scope.autoRetry.enableMaxAttempts = false
 
+  $scope.methods = Object.assign({}, defaultHttpMethods)
+  $scope.httpMethodKeys = Object.keys(defaultHttpMethods)
```
bausmeier: This isn't necessary. You can loop through the keys with `ng-repeat`.

PR #455, opened 2017-11-13
Ohm 378 limit transactions

andrew-w-ross:
Biggest issue is really adding dom nodes indefinitely has been fixed but also made sure the actual transactions don't get added.


app/views/transactions.html
```diff
@@ -105,7 +94,7 @@ <h3 style="margin: 20px;" ng-if="transactionsSelected.length < defaultBulkRerunL
                 </tr>
               </thead>
               <tbody>
-                <tr class="table-list" ng-repeat="transaction in transactions" ng-class="{transactionsNew: baseIndex + $index < 0}" ng-click='viewTransactionDetails("transactions/" + transaction._id, $event)'>
+                <tr class="table-list" ng-repeat="transaction in transactions|limitTo:settings.filter.limit" ng-class="{transactionsNew: baseIndex + $index < 0}" ng-click='viewTransactionDetails("transactions/" + transaction._id, $event)'>
```
bausmeier: This limit breaks the 'Show More' functionality.

andrew-w-ross: Good catch

PR #461, opened 2017-11-15
OHM-543: updated rest service url to include path

nthfloor2:
Added new config parameter and updated rest service to check for this path and append it to the server url.

### DIFF
--- a/app/config/default.json
+++ b/app/config/default.json
@@ -3,7 +3,7 @@
   "minimumCoreVersion": "4.0.0-rc.4",
   "protocol": "https",
   "host": "localhost",
-  "host_path": "",
+  "hostPath": "",
   "port": 8080,
   "title": "Admin Console",
   "footerTitle": "OpenHIM Administration Console",
--- a/app/config/default.json
+++ b/app/config/default.json
@@ -3,6 +3,7 @@
   "minimumCoreVersion": "4.0.0-rc.4",
   "protocol": "https",
   "host": "localhost",
+  "host_path": "",
   "port": 8080,
   "title": "Admin Console",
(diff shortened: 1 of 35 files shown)


nthfloor2:
Had to update config property for each test file - we should rethink this approach to rather pull config settings from a central location instead of defining it for each file.

app/config/default.json
```diff
@@ -3,6 +3,7 @@
   "minimumCoreVersion": "4.0.0-rc.4",
   "protocol": "https",
   "host": "localhost",
+  "host_path": "",
```
bausmeier: @nthfloor2 All of the other keys are named using camel case. We should stick to that convention for this one as well.

app/scripts/services/rest.js
```diff
@@ -2,8 +2,9 @@ export function Api ($rootScope, $resource, config) {
   // fetch API server details
   let protocol = config.protocol
```
andrew-w-ross: Change let to const.

app/scripts/services/rest.js
```diff
@@ -2,8 +2,9 @@ export function Api ($rootScope, $resource, config) {
...
   let host = config.host
+  let hostPath = config.hostPath.startsWith('/') ? config.hostPath.replace(/\/$/, '') : '/' + config.hostPath.replace(/\/$/, '')
```
andrew-w-ross: Can we just use a single regex for this?

nthfloor2: @andrew-w-ross not sure how you would do this using regex, but there is a nifty string library which can do this check and replace, so have included and used it instead.

PR #485, opened 2018-11-06
disable additional minification config

BMartinos:
to ensure bundle.js is built correctly with all the relevant code included.

Before this change, the graphs did not have any lables present when viewing a production bundle.js

OHM-650

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,21 +11,28 @@
 			"dev": true
 		},
 		"@types/connect": {
-			"version": "3.4.31",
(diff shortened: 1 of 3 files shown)


BMartinos:
To test this fix, You will need to view the `dist/index.html` directly in your browser. Running `npm run build` and `npm run build:prod` produced a different result for the production bundle.js

BMartinos:
Updating the webpack dependency (not latest version though) has fixed various vulnerabilities within the project. Before the change, there were about **250+ vulnerabilities** and after the fix is applied, there is only **17 vulnerabilities**

BMartinos:
@MattyJ007 the merge conflict has been resolved

PR #491, opened 2018-12-13
Ohm 670 update travis builds

bradsawadye:


.travis.yml
```diff
@@ -5,3 +5,5 @@ after_success:
 node_js:
   - '8.6'
 sudo: false
+notifications:
+  slack: jembihealthsystems:mlQYVFbijxcZkesCt7G5VBoM
```
BMartinos: Lets add the on_success/on_failure settings to these notifications as well. 

Lets get notified everytime the build fails, but only get notified on success if the status has been updated.
This means that if the build is in a passing status and we add more commits that also pass the build, we dont get those notifications each time, only when the build fails will we be notified

https://docs.travis-ci.com/user/notifications#configuring-slack-notifications
```
notifications:
  slack:
    rooms:
      - <account>:<token>#development
      - <account>:<token>#general
    on_success: change # default: change
    on_failure: always # default: always
```

PR #496, opened 2019-01-10
set the minimum body retention age to 1

bradsawadye:


bausmeier:
@bradsawadye @MattyJ007 Why was this change made? The validation should already have been done here: https://github.com/jembi/openhim-console/blob/5aa0069c4847d9662dc1307ef52f44ddccd88a7f/app/scripts/controllers/channelsmodal.js#L218-L222

If it wasn't working then I think that we need to figure out why. We don't want to be using HTML5 native validation since that's not consistent with the rest of our validation and does not display errors on the tabs, etc.

PR #504, opened 2019-08-27
MED-1368 update dependencies

MattyJ007:


app/scripts/utils/utils.js
```diff
@@ -87,7 +87,7 @@ export function isCoreVersionCompatible (minVersion, actualVersion) {
   const v1 = minVersion.split('.')
   const v2 = actualVersion.split('.')
 
-  if (v1[0] === v2[0]) {
+  if (v1[0] <= v2[0]) {
```
BMartinos: Would this cater for resolving of the Bug logged within Jira? or is this till a bigger issue that needs resolving?
https://jembiprojects.jira.com/browse/OHM-741

app/config/default.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.13.0-rc.1",
+  "version": "1.13.3",
```
BMartinos: Should this release not be `1.13.1` instead? we seem to be jumping 3 patch versions here

PR #508, opened 2019-10-17
OHM-900 Fix completed status color

MattyJ007:


app/views/transactions.html
```diff
@@ -119,7 +119,7 @@ <h3 style="margin: 20px;" ng-if="transactionsSelected.length < defaultBulkRerunL
                   <td data-title="Client">{{ clientsIdNameMap[transaction.clientID] }}</td>
                   <td data-title="Status">
                     <span ng-if="transaction.status === 'Successful'" class="label label-success">Successful</span>
-                    <span ng-if="transaction.status === 'Completed'" class="label label-success">Completed</span>
+                    <span ng-if="transaction.status === 'Completed'" class="label label-warning">Completed</span>
```
BMartinos: hmmm looking at this again. The `warning` class seems to make sense here initially to differentiate between `successful` and `completed` but we also have a label for `completed with errors` which will have the same colour code. Maybe we can create a new class for the `Completed with errors` label to differentiate between the two? A color between red and yellow maybe

PR #510, opened 2019-11-07
TRACE-37: Add packaging script for building bundle release version of the console

BMartinos:
TRACE-37

### DIFF
--- a/packaging/README.md
+++ b/packaging/README.md
@@ -13,9 +13,9 @@ PPA: release
 
 ## Bundled Release
 
-A bundled release will ensure all the relevant dependencies are downloaded and bundled into a built version of the OpenHIM console. Only the relevant scripts needed to run the OpenHIM core is added to the bundled release.
+A bundled release will ensure all the relevant dependencies are downloaded and bundled into a built version of the OpenHIM console. Only the relevant scripts needed to run the OpenHIM console are added to the bundled release.
 
-To create a new build release execute to the below command. This does assume that your Linux distribution has the `zip` and `tar` modules installed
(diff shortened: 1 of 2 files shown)


packaging/README.md
```diff
@@ -11,6 +11,18 @@ You must also have an environment variable set with the id of the key to use. Vi
 Login: openhie  
 PPA: release
 
+## Bundled Release
+
+A bundled release will ensure all the relevant dependencies are downloaded and bundled into a built version of the OpenHIM console. Only the relevant scripts needed to run the OpenHIM core is added to the bundled release.
+
```
bradsawadye: This should be - 

> Only the relevant scripts needed to run the OpenHIM console are added to the bundled release

packaging/README.md
```diff
@@ -11,6 +11,18 @@ You must also have an environment variable set with the id of the key to use. Vi
...
+To create a new build release execute to the below command. This does assume that your Linux distribution has the `zip` and `tar` modules installed
```
bradsawadye: this should be 

> execute the below command

packaging/build-release-zip.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
```
bradsawadye: I think we should add the 'set -eu' command, so that when there is a failure, errors are not ignored, and we bundle the right code

BMartinos: done in c8ff75b

packaging/README.md
BMartinos: done in 2642044

packaging/README.md
BMartinos: done in 2642044

PR #527, opened 2020-08-27
Ohm 1031 deps update

BMartinos:
Updated Project package dependencies

To keep the modules up to date and secure


MattyJ007:
These changes are covered in https://github.com/jembi/openhim-docs/pull/42

PR #531, opened 2020-10-21
Ohm 1038 refactor transaction sections

bradsawadye:


app/scripts/services/rest.js
```diff
@@ -36,6 +36,22 @@ export function Api ($rootScope, $resource, config) {
 
     Transactions: $resource(server + '/transactions/:transactionId', { transactionId: '@_id' }),
 
+    TransactionBodies: function(transactionId, bodyId, start, end) {
+      const headers = {}
+      if (
+        end &&
```
rcrichton, line 42: I think this should be `start && end && end > start` right?

```suggestion
        start &&
```

rcrichton (changes requested):
@bradsawadye Looks very good! Nice work. Just a small comment.

Also, I don't see any tests. It would be great to get some coverage over these new additions. There are some existing tests for the controller which would help guide you. Let me know if you have any issues or if these are too difficult to test for some reason.

rcrichton:
Another note, on large bodies (> 1Mb) the console tab seems to freeze. This is probably due to the beautification or highlighting. I'm not sure if this was a problem before this work or not.

rcrichton:
> Another note, on large bodies (> 1Mb) the console tab seems to freeze. This is probably due to the beautification or highlighting. I'm not sure if this was a problem before this work or not.

Ignore this, I had the wrong core version.

app/scripts/controllers/transactionDetails.js
```diff
@@ -232,14 +256,16 @@ export function TransactionDetailsCtrl ($scope, $uibModal, $compile, $location,
...
+  $scope.viewBodyDetails = function (type, content, headers, bodyId) {
     $uibModal.open({
       template: transactionsBodyModal,
       controller: TransactionsBodyModalCtrl,
       windowClass: 'modal-fullview',
       resolve: {
         bodyData: function () {
-          return { type: type, content: content, headers: headers }
+          return {
+            type: type, content: content, headers: headers, transactionId: $routeParams.transactionId, bodyId
```
rcrichton, line 267: You will also need to pass the current body params (`start, end, bodyLength` and a partial body flag) to the modal. the model need this for future requests that it might make for the response body.

app/scripts/controllers/transactionsBodyModal.js
```diff
@@ -1,16 +1,49 @@
-import { beautifyIndent, returnContentType } from '../utils'
+import { beautifyIndent, returnContentType, retrieveBodyProperties } from '../utils'
+import { Api } from '../services/rest'
 
-export function TransactionsBodyModalCtrl ($scope, $uibModalInstance, bodyData) {
+export function TransactionsBodyModalCtrl ($scope, $uibModalInstance, config, Alerting, bodyData) {
   $scope.bodyData = bodyData
+  const defaultLengthOfBodyToDisplay = config.defaultLengthOfBodyToDisplay
 
+  $scope.partialBody = false
```
rcrichton, line 8: Following on from above, this can't be defaulted to false as at this point as the request has already been executed outside of the modal and the modal need to be passed this information.

rcrichton (changes requested):
Just some additional comments while testing this out.

app/scripts/controllers/transactionsBodyModal.js
```diff
@@ -1,16 +1,49 @@
...
+
+        if (bodyData.headers && returnContentType(bodyData.headers)) {
+          let bodyTransform = beautifyIndent(returnContentType(bodyData.headers), response.data)
+          $scope.bodyData.content = bodyTransform.content
+        }
+      }).catch(err => {
+        Alerting.AlertAddServerMsg(err.status)
+      })
+    }
+    $scope.retrieveBodyData()
```
rcrichton, line 44: Since we already have the body data we don't need to call this here. It can be called later when we need to laod more of the body.

app/scripts/controllers/transactionsBodyModal.js
```diff
@@ -1,16 +1,49 @@
...
     if (bodyData.headers && returnContentType(bodyData.headers)) {
       const bodyTransform = beautifyIndent(returnContentType(bodyData.headers), bodyData.content)
       $scope.bodyData.content = bodyTransform.content
     }
   }
 
+  if (
+    $scope.bodyData &&
+    $scope.bodyData.transactionId &&
+    $scope.bodyData.bodyId
```
rcrichton, line 20: This ends up being undefined currently.

app/scripts/controllers/transactionsBodyModal.js
```diff
@@ -1,16 +1,49 @@
-import { beautifyIndent, returnContentType } from '../utils'
+import { beautifyIndent, returnContentType, retrieveBodyProperties } from '../utils'
+import { Api } from '../services/rest'
 
-export function TransactionsBodyModalCtrl ($scope, $uibModalInstance, bodyData) {
+export function TransactionsBodyModalCtrl ($scope, $uibModalInstance, config, Alerting, bodyData) {
```
rcrichton, line 4: Need to import `API` here to.

PR #550, opened 2021-09-09
OHM-1085 Update console validation for audit log patientID filter

michaelloosen:
To fix a bug allowing the capturing of non-alphanumeric characters in the audit log patientID filter field as well as correct a bug with the date filters.
OHM-1085 OHM-927


MattyJ007 (approved):
Great work 💯

PR #558, opened 2021-09-20
Ohm 1052 reduce docker image size

Ken473:


.dockerignore
```diff
@@ -1 +1,4 @@
-node_modules
\ No newline at end of file
+node_modules
+infrastructure
+packaging
+test
```
MattyJ007, line 4: Please add a newline to the end of the file

Dockerfile
```diff
@@ -1,23 +1,18 @@
 # Openhim Console Dockerfile for latest changes
-FROM ubuntu:16.04
+FROM node:14.17-alpine
```
MattyJ007, line 2: This should be nginx:alpine

Dockerfile
```diff
@@ -1,23 +1,18 @@
...
+RUN apk upgrade --update-cache --available
+RUN apk add git
```
MattyJ007, line 4: Why is this here?

Dockerfile
```diff
@@ -1,23 +1,18 @@
...
-RUN git clone https://github.com/jembi/openhim-console.git
-
-WORKDIR /etc/openhim-console
+COPY package.json /app
+COPY . .
 
 # Install dependencies and build
-RUN npm i
+RUN npm install
+RUN npm install webpack-dev-server && npm install webpack webpack-cli --save-dev
```
MattyJ007, line 14: These are dev dependencies, are they needed for the prod build?

infrastructure/docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
 
   openhim-core:
     container_name: openhim-core
-    image: jembi/openhim-core:latest
+    image: openhim-alpinev3-multi-staged
```
MattyJ007, line 17: This is not a valid image. Please revert

infrastructure/docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
 
   openhim-console:
     container_name: openhim-console
-    image: jembi/openhim-console:latest
+    image: openhim-console-alpinev1
```
MattyJ007, line 37: This is not a valid image. Please revert

Dockerfile
```diff
@@ -1,23 +1,18 @@
...
-RUN apt-get install -y bzip2
-RUN apt-get install -y curl
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
-RUN apt-get install -y nodejs
-
-# Clone Openhim-console repo
-RUN git clone https://github.com/jembi/openhim-console.git
-
-WORKDIR /etc/openhim-console
+COPY package.json /app
```
MattyJ007, line 9: Nginx will serve html files in this dir `/usr/share/nginx/html`

MattyJ007 (changes requested):

Dockerfile
Ken473: I found some dependencies in the package.json file that referenced git repositories. Without these the application would not build.

Dockerfile
Ken473: duly noted. Will change these

Dockerfile
MattyJ007: I see, so then I guess perhaps the node alpine can be part of the multi-stage build. Then just copy over the `/dist` directory into the alpine image

Dockerfile
MattyJ007: I see they are needed but the regular install will include them

Dockerfile
```diff
@@ -1,23 +1,18 @@
...
+COPY . .
 
 # Install dependencies and build
-RUN npm i
+RUN npm install
+RUN npm install webpack-dev-server && npm install webpack webpack-cli --save-dev
+
```
MattyJ007, line 15: The two lines below can be removed since `npm install` covers *npm build* (it automatically triggers the `npm run prepare` script).
 and nginx will serve the HTML without the http-server

MattyJ007 (changes requested):

infrastructure/docker-compose.yml
Ken473: sure

infrastructure/docker-compose.yml
Ken473: sure

Dockerfile
MattyJ007: Yes, just run `npm install` and that will take care of everything

Dockerfile
Ken473: We should change the npm run build to npm run build prod. This one includes the dev dependencies and I do not know if it includes anything that may affect you in production

Dockerfile
Ken473: cool

infrastructure/docker-compose.yml
MattyJ007: This has not been reverted

infrastructure/docker-compose.yml
MattyJ007: This has not been reverted

Dockerfile
```diff
@@ -1,25 +1,18 @@
...
-RUN npm i -g http-server 
+RUN npm install
+
+FROM nginx:mainline-alpine
+WORKDIR /usr/share/nginx/html
 
-# Host and run server
-CMD http-server ./dist -p 80
\ No newline at end of file
+COPY --from=build /app/dist  /usr/share/nginx/html
```
MattyJ007, line 18: Add new line to end of file

Dockerfile
MattyJ007: Why is this step separate from `COPY . .` ?

Dockerfile
```diff
@@ -1,25 +1,18 @@
 # Openhim Console Dockerfile for latest changes
-FROM ubuntu:16.04
+FROM node:14.17-alpine as build
+RUN apk upgrade --update-cache --available
```
MattyJ007, line 3: Is this step necessary?

MattyJ007 (changes requested):
Looks good - just a few minor comments

Dockerfile
Ken473: That is done

Dockerfile
Ken473: I had added it to make sure we are not missing any required dependencies. Let me test and see what's happening

Dockerfile
Ken473: forgot to remove it. Should be gone now.

Dockerfile
Ken473: Looks good so far.

infrastructure/docker-compose.yml
Ken473: Please check new commit

infrastructure/docker-compose.yml
Ken473: same as above

MattyJ007 (approved):

PR #579, opened 2022-09-09
Plat 249 remove limit on bulk rerun

bradsawadye:


app/scripts/controllers/transactions.js
```diff
@@ -413,18 +413,23 @@ export function TransactionsCtrl ($scope, $uibModal, $location, $timeout, $inter
 
   const refreshSuccess = function (transactions) {
     // on success
-    $scope.transactions = transactions
+    if (transactions.count) {
```
marrouchi, line 416:
```suggestion
    if (transactions.length) {
```
Is transactions a object or an array ?

app/scripts/controllers/transactions.js
```diff
@@ -413,18 +413,23 @@ export function TransactionsCtrl ($scope, $uibModal, $location, $timeout, $inter
...
+      $scope.transactions = []
+      $scope.transactionsCount = transactions.count
+    } else {
+      $scope.transactions = transactions
```
marrouchi, line 420: `transactionsCount` should be updated here as well right ?

marrouchi (changes requested):
I don't have background on the intent of the PR, so I did a review with the best of my knowledge :)

app/scripts/controllers/transactions.js
bradsawadye: The endpoint can return an array of transactions or an object containing the count property, which is the length of the transactions to be rerun (for bulk reruns).

marrouchi: Thanks for the clarification

marrouchi (approved):

michaelloosen (approved):
LGTM

PR #599, opened 2023-02-16
Plat 636 adjust login keycloak

nour-borgi:
In this PR added keycloak integration

To test this PR, you can follow these steps:
- Pull [this branch](https://github.com/jembi/platform/pull/221) of Openhim-keycloak in platform and enable openhim sso by adding this to .env.loacl: `KC_OPENHIM_SSO_ENABLED=true`
- Build the image and start keycloak only: `./instant-linux package init -n identity-access-manager-keycloak  --only --dev --env-file=.env.local`
- Start openhim core and openhim console (using the keycloak branch as well)
- To enable keycloak SSO in openhim console, you need to set "ssoEnabled" to true in the app/config/default.json.


app/views/users.html
```diff
@@ -65,6 +66,7 @@ <h2><i class="glyphicon glyphicon-user"></i>&nbsp;&nbsp;Users</h2>
                       <span ng-if="user.user.locked" style="color: #d9534f" tooltip-html-unsafe="This user is currently locked<br />They will need to set their password or admin needs to set it for them"><i class="glyphicon glyphicon-lock"></i></span>
                     </td>
                     <td data-title="Name">{{ user.user.firstname }} {{ user.user.surname }}</td>
+                    <td data-title="Name">{{ user.user.provider }}</td>
```
bradsawadye, line 69: `data-title` should be different. "Name" is already taken

nour-borgi: nice done! thanks for the review

app/scripts/controllers/login.js
```diff
@@ -172,4 +209,17 @@ export function LoginCtrl ($scope, login, $window, $location, $timeout, $rootSco
...
+    if (config.ssoEnabled) {
+      keycloak.keycloakInstance.init({
+        onLoad: 'login-required',
+        // Must match to the configured value in keycloak
+        redirectUri: $window.location.origin,
+        checkLoginIframe: false
+      })
+    } else {
+      Alerting.AlertAddMsg('login', 'danger', 'Sign-in with Keycloak is disabled. Please contact the server administrator.')
+    }
```
bradsawadye, line 223: The logic in the else statement here will never be invoked because the function `signInWithKeyCloak` is only ever run when the `config.ssoenabled` is true. The button `Sign-In with keycloak` is in a div that has the ng-show directive which depends on `config.ssoenabled`.

nour-borgi: yes I added that to double make sure this will not be triggered even if there are changes in the views, 
if you think it is not necessary I can remove it

bradsawadye: i dont think its necessary. Its code that will never be run. Lets remove it

nour-borgi: Cool done :)

app/template.html
```diff
@@ -45,7 +45,7 @@ <h1>
                       <li class="dropdown">
                         <a href="" class="dropdown-toggle" data-toggle="dropdown"><i class="glyphicon glyphicon-user"></i><span ng-bind="sessionUser"></span><b class="caret"></b></a>
                         <ul class="dropdown-menu animated fadeInUp">
-                          <li><a href="#!/profile">Profile</a></li>
+                          <li ng-if="sessionProvider !== 'keycloak'"><a href="#!/profile">Profile</a></li>
```
rcrichton, line 48: Hmm, there are some setting in a users profile that we still need to be able to access. Like setting their reporting preferences and filtering preferences. We will still need to view this profile page, however, we won't be able to change certain fields.

rcrichton (changes requested):
Looking good, however I haven't functionally tested yet. What is an easy way to set this all up with keycloak for testing?

nour-borgi:
Hi @rcrichton, I added in the description how to test this PR.

app/template.html
nour-borgi: cool done :)

rcrichton (approved):

rcrichton:
Looks awesome and works well. Well done :)

PR #610, opened 2023-05-19
feat: add docker file for Google Cloud Run

marrouchi:
# Motivation

Google Cloud Run does not offer a way to mount volumes nor configs, as it supports only stateless containers.

This PR adds a custom Dockerfile for stateless containers when it does env var substitution at the entrypoint level.


michaelloosen (dismissed):

infrastructure/google-cloud-run/start-nginx.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+export EXISTING_VARS=$(printenv | awk -F= '{print $1}' | sed 's/^/\$/g' | paste -sd,);
+
+cat config/default-env.json | envsubst $EXISTING_VARS | tee config/default.json
```
rcrichton, line 4: If you are selecting all var, isn't that the same as running `envsubst` with no parameters?

app/config/default-env.json
rcrichton, line 1: Should we rather have env vars for all option in this file. The shell script might have to ensure the env vars are set with a default value where needed. i.e. `export PROTOCOL=${PROTOCOL:-https}` or something.

rcrichton (changes requested):
With the above changes, I see no reason why this couldn't be the default docker image as it would be fully backwards compatible. Maybe we should consider that?

app/config/default-env.json
```diff
@@ -0,0 +1,19 @@
...
+  "hostPath": "",
+  "port": "${OPENHIM_MEDIATOR_API_PORT}",
+  "title": "Admin Console",
+  "footerTitle": "OpenHIM Administration Console",
+  "footerPoweredBy": "<a href='http://openhim.org/' target='_blank'>Powered by OpenHIM</a>",
+  "loginBanner": "",
+  "mediatorLastHeartbeatWarningSeconds": 60,
+  "mediatorLastHeartbeatDangerSeconds": 120,
+  "showLoginForm": true,
+  "ssoEnabled": ${KC_OPENHIM_SSO_ENABLED},
```
marrouchi, line 15:
```suggestion
  "ssoEnabled": ${KC_OPENHIM_SSO_ENABLED},
```
@michaelloosen  in platform this value is within double quotes, as string it's evaluated to "true".

arran-standish (approved):
I like the idea of updating the base image instead as this simplifies platform, as currently it is maintaining a default.json file that it does envsubst on. Which means we have the same file duplicated across repo's that we need to keep in sync. With this change we can remove all that from platform and just pass in the relevant environment variables

michaelloosen (approved):

PR #617, opened 2023-09-13
TB-173 show transaction error with no response

arran-standish:
With the changes to Openhim in 8.1.2 it is possible for a transaction to be in a failed state but without receiving a response back from the down stream service (i.e.: Openhim crashes midflight) but it would still be nice to be able to see the error message on the now failed transaction


MatthewErispe (approved):
@arran-standish LGTM

PR #619, opened 2023-11-14
Create a new admin page for managing apps

medchedli:
This pull request moves the Add/Edit/Delete user actions of OpenHIM Portal to a dedicated admin page. This change will make it easier for administrators to manage OpenHIM Apps.

Changes:
- Move the Add/Edit/Delete actions to a dedicated admin page.

- Updated the UI to reflect the changes.

Benefits:

- Improves the overall user experience for both admins and regular users


packages/openhim-core-api/.prettierignore
```diff
@@ -0,0 +1,8 @@
+.gitignore
+.prettierignore
+yarn.lock
+yarn-error.log
+package-lock.json
+dist
+coverage
+pnpm-lock.yaml
```
bradsawadye, line 8:
```suggestion
pnpm-lock.yaml

```

packages/openhim-core-api/src/jembi-openhim-core-api.ts
```diff
@@ -0,0 +1,124 @@
...
+    )
+    const appsGroupedByCategory = categories.map(category => {
+      const apps = response.data.filter(app => app.category === category)
+      return {category, apps}
+    })
+    return appsGroupedByCategory
+  } catch (error) {
+    throw error
+  }
+}
```
bradsawadye, line 123: try and catch blocks are redundant here. We are catching the error and then throwing it without modifying it

packages/portal-admin/.prettierignore
```diff
@@ -0,0 +1,8 @@
+.gitignore
+.prettierignore
+yarn.lock
+yarn-error.log
+package-lock.json
+dist
+coverage
+pnpm-lock.yaml
```
bradsawadye, line 8:
```suggestion
pnpm-lock.yaml

```

brett-onions (changes requested):

brett-onions (approved):
LGTM

packages/openhim-core-api/src/jembi-openhim-core-api.ts
```diff
@@ -0,0 +1,84 @@
...
+/* ****************************************************************************************************************
+if running locally set REACT_APP_OPENHIM_API_BASE_URL environment variable to point to OpenHIM Core API base URL
+ **************************************************************************************************************** */
+
+if (!process.env.REACT_APP_OPENHIM_API_BASE_URL) {
+  throw new Error('REACT_APP_OPENHIM_API_BASE_URL is not set')
+}
+
+const API_URL =
+  process.env.REACT_APP_OPENHIM_API_BASE_URL || 'https://localhost:8080/'
```
drizzentic, line 12: Shouldn't this be able to sort the check in L7? we already setting a default value if we can't get the environment variable. Otherwise, if we want to enforce the check the default URL will never be used.

medchedli: Yes, you are absolutely right, I fixed this issue yesterday locally, and I will push the fix with the next commit

packages/openhim-core-api/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@jembi/openhim-core",
```
Lumiere-Lodi, line 2: Fixing the package name to match what we have in root-config index.ejs

```suggestion
  "name": "@jembi/openhim-core-api",
```

Lumiere-Lodi (commented):
one suggestion left

PR #621, opened 2024-01-22
General Improvements

Lumiere-Lodi:
This PR: 

**1. Add: Global theme**

 a global theme that can be used by any react module in the micro frontend application. 

The theme is called using `import theme from '@jembi/openhim-theme'` and make sure you add the module to the declarations.d.ts file using `declare module '@jembi/openhim-theme'` 

When run successfully you will access the global theme style. 

**2. Refactoring: Portal admin Module**
- Fixed state drilling issue with a better flow 
- Fix unnecessary rerender 
- Restructured component layout in react to avoid HOC issues. 
- Redesigned form dialog

**3. Fix: Sidebar**
- Fixed theme collusion between legacy app and micro frontend. Ratio between Pixel and rem is causing a problem. Solution was to use Typography instead of ListItemText. With Typography we can set a fix fontsize pixel size that will be used across all modules that calls the sidebar.


medchedli (commented):
:+1:

PR #627, opened 2024-04-25
Allow for beautification of custom json header

BMartinos:
Allow requests which contain custom application header for json to be parsed and displayed in a nice format This assist for cases for example, FHIR data "application/fhir+json".


drizzentic (approved):
LGTM!

PR #639, opened 2024-07-02
ft: Creating Client App

brett-onions:


packages/clients-app/src/root.component.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { render } from "@testing-library/react";
+import Root from "./root.component";
+
+describe("Root component", () => {
+  // it("should be in the document", () => {
+  //   const { getByText } = render(<Root name="Testapp" />);
+  //   expect(getByText(/Testapp is mounted!/i)).toBeInTheDocument();
+  // });
+});
```
drizzentic, line 9: Consider removing this for now till we start implementing test.

packages/openhim-core-api/src/jembi-openhim-core-api.ts
```diff
@@ -114,17 +114,91 @@ export async function getImportMap(): Promise<any> {
   return response.data
 }
 
+export async function fetchRoles(): Promise<any> {
+  await ensureApiClientInitialized()
+  const response = await apiClient.get('/roles')
+  return response.data
+}
```
drizzentic, line 121: Based on the discussion we had for client role. We should probably be fetching this from clients and not the roles endpoint.

drizzentic (commented):
LGTM! Just need to do cleanups for the unused files and modify the roles method.

marvinIsSacul (approved):
Looks good.

Just need to fix the failing CI

packages/clients-app/src/root.component.test.tsx
brett-onions: When running the npm test command with the --passWithNoTests flag it still fails. So I will just leave this place holder to prevent failures.

drizzentic: It will still fail. So it's best to delete the file all together

PR #669, opened 2024-09-06
feat: use hashRouter

marvinIsSacul:
fix: addAutoRewriteRules

### DIFF
--- a/packages/root-config/src/microfrontend-layout.html
+++ b/packages/root-config/src/microfrontend-layout.html
@@ -1,38 +1,23 @@
 <single-spa-router>
   <main>
     <application name="@jembi/openhim-header"></application>
-    <div style="padding: 0 5%; background-color: #f1f1f1; min-height: auto;">
+    <div style="padding: 0 5%; background-color: #f1f1f1; min-height: auto">
       <route default>
-        <route path="#!" exact="true">
-        </route>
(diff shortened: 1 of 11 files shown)


packages/root-config/src/microfrontend-layout.html
```diff
@@ -1,38 +1,23 @@
...
+        <route path="#!/groups"> </route>
+        <route path="#!/mediators"> </route>
+        <route path="#!/certificates"> </route>
+        <route path="#!/export-import"> </route>
+        <route path="#!/logs"> </route>
+        <route path="#!/about"> </route>
+        <route path="#!/profile"> </route>
+        <route path="#!/portal-admin"> </route>
+        <div style="flex-grow: 1">
+          <application name="@jembi/legacy-console"></application>
```
drizzentic, line 20: which formatter are you using?

drizzentic (approved):
LGTM! Just some minor formatting issues

packages/root-config/src/microfrontend-layout.html
marvinIsSacul: I'm using prettier

marvinIsSacul: Should I undo them ?

PR #682, opened 2024-09-17
Cu 86c0835pw add a loaderspinner when the request is being processed before loading of datagrid data

brett-onions:


rcrichton:
Task linked: [CU-86c0835pw Add a loader/spinner when the request is being processed before loading of datagrid data](https://app.clickup.com/t/86c0835pw)

packages/transaction-log/src/components/common/animated.table.row.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const AnimatedTableRow = ({
+  children,
+  initialColor,
+  finalColor,
+  onClick
+}) => {
+  const ref = useRef(null)
+  const [completed, setCompleted] = useState(false)
+
+  useEffect(() => {
```
marvinIsSacul, line 15: Does this cause an infinite loop of some sort? 🤔

`useEffect` runs whenever `completed` is modified. Yet `useEffect` also causes `completed` to be modified.

packages/transaction-log/package.json
```diff
@@ -60,6 +60,7 @@
     "@types/systemjs": "^6.1.1",
     "@types/webpack-env": "^1.16.2",
     "axios": "^1.7.2",
+    "moment": "^2.30.1",
```
marvinIsSacul, line 63: moment.js appears to be a bit outdated. Can you check out date-fns?

This is optional.

brett-onions: I used moment because it was the Lib that the previous version of the OpenHIM used. I can migrate over to date-fns since on the moment.js page it says it is in legacy mode.

packages/transaction-log/src/components/common/animated.table.row.component.tsx
brett-onions: Thanks, an infinite loop is not occuring but I removed it just to be safe, and the behaviour is still the same.

marvinIsSacul (approved):
LGTM!

PR #698, opened 2024-10-08
Implement restore default on customize dialog

ItsMurumba:


packages/transaction-log/src/components/common/transactionlog.datatable.component.tsx
```diff
@@ -92,63 +92,101 @@ const TransactionLogTable: React.FC<{
   }
 
   return (
-    <Box sx={{padding: '16px'}}>
+    <Box>
```
brett-onions, line 95: could this <Box> be removed?

brett-onions (changes requested):

packages/transaction-log/src/interfaces/index.interface.ts
```diff
@@ -65,10 +65,26 @@ export interface CustomizeDialogProps {
...
+    path: boolean
+    param: boolean
+    client: boolean
+    method: boolean
+  }
   handleFilterVisibilityChange: (
     event: React.ChangeEvent<HTMLInputElement>
   ) => void
+  onRestoreDefaults: () => void
+  isDefaultState: boolean
```
marvinIsSacul, line 87:
```suggestion
  isDefaultState?: boolean
```

marvinIsSacul (approved):

packages/transaction-log/src/components/common/transactionlog.datatable.component.tsx
ItsMurumba: this is already removed @brett-onions

packages/transaction-log/src/interfaces/index.interface.ts
ItsMurumba: Why would you want to make the property optional? @marvinIsSacul

brett-onions (approved):

